PR #24, opened 2015-01-13
Add support for saving binary data in data vault

maffoo:
We've discussed the possibility of having the data vault store data in binary form, rather than as csv files. This should reduce the file size of typical data sets, and it should also make loading and saving data much faster, because we can avoid the overhead of csv encoding/decoding. I've also used mmap for the in-memory representation of the data, which will use the os-level caching of memory pages, rather than the ad-hoc memory cache we implemented for python lists and numpy arrays using DATA_TIMEOUT. I want to add tests and some performance benchmarks, but wanted to get people's comments on this prototype.

### DIFF
--- a/data_vault.py
+++ b/data_vault.py
@@ -119,7 +119,8 @@ class DataVault(LabradServer):
         # start in the root session
         c['path'] = ['']
         # start listening to the root session
-        self.session_store.get(['']).listeners.add(c.ID)
+        c['session'] = self.session_store.get([''])
+        c['session'].listeners.add(c.ID)
 
     def expireContext(self, c):
         """Stop sending any signals to this context."""
@@ -135,14 +136,13 @@ class DataVault(LabradServer):
 
     def getSession(self, c):
         """Get a session object for the current path."""
-        return self.session_store.get(c['path'])
+        return c['session']
 
     def getDataset(self, c):
         """Get a dataset object for the current dataset."""
         if 'dataset' not in c:
             raise errors.NoDatasetError()
-        session = self.getSession(c)
-        return session.openDataset(c['dataset'])
+        return c['datasetObj']
 
     # session signals
     onNewDir = Signal(543617, 'signal: new dir', 's')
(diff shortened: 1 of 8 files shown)


ejeffrey:
Two comments:
1) I am not sure about the use of mmap here.  One problem with mmap() is that you can't resize when adding new data.  This seems like it could be a performance problem on the very common case of the grapher updating a plot while the dataset is being written to.  Every add() will cause the map to be destroyed and recreated.
2) The binary format used here is very simple, and may be somewhat limiting in the future.  We could use a standard data format such as HDF5 instead would allow for future extensibility to other datatypes (integers, strings) without adding yet another data format.  HDF5 also allows multiple datasets per file (it has a full path tree structure if you want to use it), so we can put one HDF5 / folder and reduce the amount of time it takes to copy / backup the data vault.

maffoo:
Re: mmap, it's a good point about not resizing in the current implementation; I left myself a TODO to look into mremap, but I don't know how expensive mremap itself is. Note that it is at least lazy in the sense that an `add` closes the map but doesn't recreate anything until there's a request to read data, e.g. if by a grapher plotting the data live. But watching data come in live is something we do commonly, so I'd need to show that the performance is acceptable and I haven't done that.

Re data formats: my thinking was that this would be easily generalizable into rows that consist of any fixed-width labrad types, basically anything but lists or strings (unless we were to allow fixed-width versions of those. Then it's not really a new format, it's basically just a labrad list on disk, with each list element being one row. However, HDF5 could certainly be useful and would have plenty of other advantages, especially if the drivers handle all the mmap'ing themselves, which seems likely. I'll look into that.

ejeffrey:
Anyway, the code looks fine but I haven't tested it myself.

ejeffrey:
pyHDF5 has efficient IO drivers for multiple platforms.  It also specifies the datatypes internally, so there is no chance of mixing up types.

maffoo:
The refactor/consolidation was merged as part of #150. Closing this so we can go for hdf5 instead of basic binary data.

PR #32, opened 2015-02-06
Removed .asarray from dac_calibration_server.py

JulianSKelly:


pomalley:
LGTM. Especially since we're about to delete the calibration server in any event.

JulianSKelly:
Merged via command line

Issue #34, opened 2015-02-11
Qubit sequencer should support receiving DAC'ified data.


pomalley:
Currently the only way to give pre-deconvolved data to the qubit sequencer is if you give it time domain data. If we're going to do deconvolutions in pyle, it should support receiving deconvolved frequency domain data as well.

maffoo:
There's really no point in keeping frequency domain data around once you've corrected for the response. I'd say that the deconvolution should just fourier transform back to the time domain as well, just as the correction code currently does. Then it wouldn't be necessary to modify the qubit sequencer at all.

pomalley:
You're right, I was confused.

However, the data returned from the calibration server are in DAC units. The qubit sequencer does not have the ability to receive DAC'ified data. (For deconvolve=false the qubit sequencer DAC'ify's the data itself; see here: https://github.com/martinisgroup/servers/blob/master/Qubit%20Server/dev/src/org/labrad/qubits/channeldata/IqDataTime.java#L25)

So this ties in with #33--where do we want to DACify the data?

maffoo:
Given that the qubit sequencer can distinguish between floating point data and integer data, it should be easy to keep the current behavior (complex data that is already deconvolved is rescaled to DAC units assuming +/-1 full scale) and add the ability to handle dacified data (integer data is passed along as-is to ghz dacs server).

pomalley:
Yes, it proved pretty easy (though this is untested): a4bab60e19046977c0ad5fa955fcc67872137b7f

ejeffrey:
I think ideally the fpga server should be responsible for knowing the DAC
resolution and the packing format and doing dacify.  We would need to send
the ECL data separately then.

Evan

DanielSank:
So you want to send +1/-1 scaled floats for I and Q to the fpga server.

pomalley:
That would be my preference, but I'm not sure exactly how this works with the dynamic range that's set aside by the calibration server.

ejeffrey:
Yes, that is my preference.  Either the qubit sequencer or the
deconvolution server should make sure we have spare dynamic range as
necessary (so that a piAmp of 1 -> 0.5 to the fpga server), but doesn't
care whether '1' is 2^13 or 2^7 or something else, and certainly doesn't
know that the high 2 bits of each 16 bit word correspond to ECL outputs.

Evan

DanielSank:
[...]

Issue #33, opened 2015-02-11
Values should be converted to DAC units in the FPGA server, not the DAC calibration server.


pomalley:
AFAICT, the deconvolution server is what converts the full scale range (-1, 1) to the DAC range (0, 2**14).

This line: https://github.com/martinisgroup/servers/blob/master/ghzdac/correction.py#L852

This should be done in the GHz FPGA server. It would be nice to e.g. be able to plot the corrected data.

DanielSank:
While I agree that the deconvolution server should not be dacifying the wave forms, I'm not convinced that the fpga server should be doing it either. If the FPGA server is to do it I would say that it should be done in the various `DAC` subclasses to make sure that different numbers of DAC bits are supported.

DanielSank:
I see that things are moving here. Do you want me to put the dacification into the DAC subclasses? I'm fine to do this as long as everyone agrees that it makes sense.

pomalley:
I'm sort of half-way through a qubit sequencer refactoring-ish thing (@maffoo helped a lot but I've been slow). We also have the DAC builds now, too. So let's talk this out; I'll be in in a minute.

DanielSank:
I'm at the Google building but we can chat after J-Dog's thesis offense.

On Mon, Feb 23, 2015 at 9:10 AM, pomalley <notifications@github.com> wrote:

> I'm sort of half-way through a qubit sequencer refactoring-ish thing (
> @maffoo <https://github.com/maffoo> helped a lot but I've been slow). We
> also have the DAC builds now, too. So let's talk this out; I'll be in in a
> minute.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/martinisgroup/servers/issues/33#issuecomment-75585848>
> .
>



-- 
Daniel Sank

JulianSKelly:
Given that I have completed my thesis defense, I think we should proceed on this front.

DanielSank:
Yeah good point. @pomalley, get your butt over here.

PR #35, opened 2015-02-12
DAC'ified data to the qubit sequencer

pomalley:
This has been tested now (well, a Rabi scan runs, at least), and I think it might be a generically useful feature in the future.

### DIFF
--- a/Qubit Server/dev/src/org/labrad/qubits/QubitContext.java
+++ b/Qubit Server/dev/src/org/labrad/qubits/QubitContext.java
@@ -57,1097 +57,1097 @@ import com.google.common.collect.Maps;
 
 public class QubitContext extends AbstractServerContext {
 
-	private Experiment expt = null;
-	private Context setupContext = null;
-
-	private Request nextRequest = null;
-	private int runIndex;
-	private Data lastData = null;
-
-	private boolean configDirty = true;
-	private boolean memDirty = true;
-	private boolean sramDirty = true;
-
-	/**
-	 * Initialize this context when it is first created.
-	 */
-	@Override
-	public void init() {
-		// this is the context in which we will create setup packets
-		// we make it different from our own context to avoid potential
-		// lockups due to making recursive calls in a context.
-		setupContext = new Context(getConnection().getId(), 1);
-	}
-
-	@Override
-	public void expire() {}
-
-
-	/**
-	 * Get the currently-defined experiment in this context.
-	 */
-	private Experiment getExperiment() {
-		Preconditions.checkNotNull(expt, "No sequence initialized in this context.");
-		return expt;
-	}
-
-	/**
-	 * Set the current experiment in this context
-	 * @param expt
-	 */
-	private void setExperiment(Experiment expt) {
-		this.expt = expt;
-	}
-
-
-	/**
-	 * Get a channel from the experiment that is of a particular Channel class
-	 * this unpacks the channel descriptor directly from the incoming LabRAD data
-	 */
-	private <T extends Channel> T getChannel(Data data, Class<T> cls) {
-		if (data.matchesType("s")) {
-			String device = data.getString();
-			return getChannel(device, cls);
-		} else if (data.matchesType("ss")) {
-			String device = data.get(0).getString();
-			String channel = data.get(1).getString();
-			return getChannel(device, channel, cls);
-		} else {
-			throw new RuntimeException("Unknown channel identifier: " + data.pretty());
-		}
-	}
-
-	/**
-	 * Get a channel from the experiment that is of a particular class
-	 */
-	private <T extends Channel> T getChannel(String device, String channel, Class<T> cls) {
-		return getExperiment().getDevice(device).getChannel(channel, cls);
-	}
-
-	/**
-	 * Get a channel from the experiment that is of a particular class.
-	 * In this case, no channel name is specified, so this will succeed
-	 * only if there is a unique channel of the appropriate type.
-	 */
-	private <T extends Channel> T getChannel(String device, Class<T> cls) {
-		return getExperiment().getDevice(device).getChannel(cls);
-	}
-
-	/**
-	 * Build a setup packet from a given set of records, using the predefined setup context
-	 * @param data
-	 */
-	private Data buildSetupPacket(String server, Data records) {
-		return Data.clusterOf(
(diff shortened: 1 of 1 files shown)


DanielSank:
I kind of think the qubit sequencer should allow sequence data in either -1/+1 units *or* dacified units, not both.

pomalley:
The ultimate plan I think is to do the dacification down in the FPGA server (as we were discussing in #33 and #34) but until we change the calibration server to _not_ return DAC units I think this is what we need.

DanielSank:
Why not change the deconvolution server instead, since that's what we want?

pomalley:
We will. We're just trying to make a series of small changes so we can make sure we don't break anything (too badly) as we go.

DanielSank:
I understand. I always worry about allowing multiple argument/return types because the more ways we purport to support doing something the more hellish it is later when we change our minds.

pomalley:
Right, I think ultimately when the JT stuff is done there's going to be a lot of code we can trash. (For example: dual block.) Part of that should probably be a pretty big refactoring of the qubit sequencer. The question is if I'll have enough energy to do it, or if I wheedle @maffoo into doing it.

Qubit Server/dev/src/org/labrad/qubits/QubitContext.java
```diff
@@ -560,7 +560,9 @@ public void sram_dual_block_delay(@Accepts("v[ns]") double delay) {
 				+ "microwave quadratures.  The length of the data should "
 				+ "match the length of the current SRAM block.  "
 				+ "An optional boolean specifies whether the data "
-				+ "should be deconvolved (default: true).")
+				+ "should be deconvolved (default: true)."
+                + "If deconvolve=false, the data can be specified as DAC-ready"
+                + "I and Q integers.")
```
maffoo: indentation is screwy here. I think the file as it stands uses tabs, and I'm guessing the new lines use spaces. Let's convert this file to use spaces everywhere.

DanielSank:
Got it. In this case I'd say the right course of action is to go ahead but post an issue so we don't forget. You can even assign it to yourself if you're feeling your oats.

Qubit Server/dev/src/org/labrad/qubits/QubitContext.java
```diff
@@ -568,11 +570,17 @@ public void sram_iq_data(@Accepts({"s", "ss"}) Data id,
 	}
 	@SettingOverload
 	public void sram_iq_data(@Accepts({"s", "ss"}) Data id,
-			@Accepts("*c") Data vals,
+			@Accepts({"*c", "(*i{I}, *i{Q})"}) Data vals,
```
maffoo: it's probably more correct to make this `*(ii)` since then there's no chance of having arrays with two different lengths. I guess that may be slightly less convenient from python, though, since that would translate into an array of clusters, and as the library currently stands, you can't really represent that efficiently.

Qubit Server/dev/src/org/labrad/qubits/channeldata/AnalogDataTime.java
```diff
@@ -8,6 +8,7 @@
 
 import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
+import sun.security.util.Length;
```
maffoo: this looks like it shouldn't be here.

pomalley:
I don't make a practice of feeling oats, usually I just cook them.

Qubit Server/dev/src/org/labrad/qubits/channeldata/IqDataTime.java
```diff
@@ -29,8 +29,18 @@ public IqDataTime(ComplexArray data, boolean isDeconvolved) {
...
+    public IqDataTime(int[] I, int[] Q) {
+        this.I = I;
+        this.Q = Q;
+        setDeconvolved(true);
+    }
+
   public void checkLength(int expected) {
-    LengthChecker.checkLengths(data.length, expected);
+    if (data == null) {
+        LengthChecker.checkLengths(I.length, expected);
```
maffoo: need to check Q as well. Nothing is forcing them to have the same length.

Qubit Server/dev/src/org/labrad/qubits/QubitContext.java
```diff
@@ -568,11 +570,17 @@ public void sram_iq_data(@Accepts({"s", "ss"}) Data id,
...
 			boolean deconvolve) {
 		IqChannel ch = getChannel(id, IqChannel.class);
-		ComplexArray c = ComplexArray.fromData(vals);
-		ch.addData(new IqDataTime(c, !deconvolve));
+        if (vals.isCluster()) {
+            Preconditions.checkArgument(!deconvolve, "Must not deconvolve if providing DAC'ified IQ data.");
```
maffoo: indentation. can't tell whether it's tabs or spaces here, but should be 2 spaces.

Qubit Server/dev/src/org/labrad/qubits/QubitContext.java
pomalley: You're right, but since this is a temporary solution (famous last words) I'd rather leave it as is because this is the return signature of the calibration server function and we're just passing the calibration server results through to this.

Qubit Server/dev/src/org/labrad/qubits/QubitContext.java
pomalley: Is 2 spaces standard for Java? Both eclipse and intellij want to go for four (iirc).

maffoo: It's configurable in your IDE. The other files here use 2, so we should go with that. Also: https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.2-block-indentation

pomalley:
I made the changes per @maffoo. I also changed tabs->spaces in QubitContext.java. The indentation is definitely 2 spaces in QubitContext, but for some reason github shows it funny. ??

maffoo:
LGTM. I would suggest keeping the tabs->spaces commit separate when you squash/rebase.

pomalley:
That was the plan. Am I supposed to but "reviewed by maffoo" on all the commits?

pomalley:
(That is, both of them.)

maffoo:
It's always nice to have the info of who did the review. For whitespace changes probably not as critical, but still a good habit to get into.

PR #38, opened 2015-02-13
Jump table bench test

pomalley:
Not for merging yet, but I'd like to get @maffoo's take on the changes to the Qubit Sequencer in the last couple commits.


GHzDACs/Cleanup/dac.py
```diff
@@ -1,6 +1,9 @@
 import numpy as np
 
-import servers.GHzDACs.Cleanup.fpga as fpga
+from ...GHzDACs import jumpTable
+from ..Cleanup import fpga
```
maffoo:
```python
from .. import jumpTable
from . import fpga
```

GHzDACs/Cleanup/dac.py
```diff
@@ -71,38 +75,37 @@
...
     # lifecycle methods
-    
+
     @inlineCallbacks
     def connect(self, name, group, de, port, board, build):
         """Establish a connection to the board."""
-        print('connecting to DAC board: %s (build #%d)'% \
-            (self.macFor(board), build))
-        
+        print('connecting to DAC board: %s (build #%d)' % \
```
maffoo: nit: line continuation not needed here.

GHzDACs/Cleanup/dac.py
```diff
@@ -125,70 +128,70 @@ def connect(self, name, group, de, port, board, build):
...
     @classmethod
     def regPing(cls):
         """Returns a numpy array of register bytes to ping DAC register"""
         regs = np.zeros(cls.REG_PACKET_LEN, dtype='<u1')
-        regs[0] = 0 # No sequence start
-        regs[1] = 1 # Readback after 2us
-        return regs 
-    
+        regs[0] = 0  # No sequence start
+        regs[1] = 1  # Readback after 2us
```
maffoo: is this two spaces before a comment thing just pycharm's doing?

GHzDACs/Cleanup/dac.py
```diff
@@ -847,50 +861,62 @@ def func():
...
             ans = found, clkinv, PHOF, tries, targetFifo
             returnValue(ans)
+
         return self.testMode(func)
-    
+
     def runBIST(self, cmd, shift, dataIn):
-        """Run a BIST on the given SRAM sequence. (DAC only)"""
+        """
+
```
maffoo: Would be nice to keep the old summary, in addition to the parameter descriptions.

GHzDACs/Cleanup/dac.py
```diff
@@ -936,8 +962,252 @@ class DacRunner_Build11(DacRunner_Build7):
...
     RUNNER_CLASS = DacRunner_Build11
 
+
 fpga.REGISTRY[('DAC', 11)] = DAC_Build11
 
+
+### Jump table
+
+class DacRunner_Build13(DacRunner_Build7):
+    # noinspection PyMissingConstructor
```
maffoo: huh?

GHzDACs/Cleanup/dac.py
```diff
@@ -936,8 +962,252 @@ class DacRunner_Build11(DacRunner_Build7):
...
+        self.sram = sram
+        self.nPackets = 0  # we don't expect any packets back
+
+    def pageable(self):
+        return False  # no paging for JT
+
+    def loadPacket(self, page, isMaster):
+        """Create pipelined load packet.  For DAC, upload mem and SRAM."""
+        if isMaster:
+            # TODO: how can we add a delay to the JT?
```
maffoo: this seems important. how do we ensure that multiple JT boards start in sync?

GHzDACs/ghz_fpga_server.py
```diff
@@ -1627,16 +1627,66 @@ def dac_run_sram(self, c, data, loop=False, blockDelay=0):
...
-            # make sure data is at least 20 words long by repeating it
-            data *= (20-1)/len(data) + 1
-        else:
-            # Set data at least 20 words long by repeating first value
+
+        if len(data) < 20:
+            # pad out to 20 with zeros
+            d2 = np.zeros(20)
+            d2[0:len(data)] = data
+            data = d2
```
maffoo: Why this change?

GHzDACs/ghz_fpga_server.py
```diff
@@ -1851,8 +1901,46 @@ def dac_bist(self, c, chan, data):
...
-    
+        # This is coming back with 64-bit ints, the coercing of which needs to be fixed in pylabrad
+        # for now we manually cast to 32-bit (long)
+        # See pylabrad github issue #43
+        returnValue((ans[0],
+                    [long(x) for x in ans[1]], [long(x) for x in ans[2]], [long(x) for x in ans[3]],
+                     ))
+
+    @setting(1301, 'DAC Bringup New', lvdsOptimize='b', lvdsSD='w', signed='b', targetFifo='w',
+             returns='*(bw)')
```
maffoo: It looks like this function returns strings. Should this be `*(bs)`?

GHzDACs/jumpTable.py
```diff
@@ -7,179 +7,201 @@
...
     def getJumpIndex(self):
         return self._jumpIndex
+
     def setJumpIndex(self, idx):
-        if JUMP_INDEX_MIN < idx and idx < JUMP_INDEX_MAX:
+        if JUMP_INDEX_MIN <= idx <= JUMP_INDEX_MAX:
             self._jumpIndex = idx
         else:
-            raise RuntimeError("Must have %s < jump index < %s"%(JUMP_INDEX_MIN, JUMP_INDEX_MAX))
+            raise RuntimeError("Must have %s < jump index < %s" % (JUMP_INDEX_MIN, JUMP_INDEX_MAX))
```
maffoo: error message says `<` but comparison is `<=`

GHzDACs/jumpTable.py
```diff
@@ -7,179 +7,201 @@
...
+
     def setJumpIndex(self, idx):
         raise RuntimeError('IDLE does not support jump table indexing')
-    
+
     def getCycles(self):
         return self._cycles
+
     def setCycles(self, cycles):
         if IDLE_MIN_CYCLES < cycles and cycles < IDLE_MAX_CYCLES:
```
maffoo: should these comparisons be `<=` too?

GHzDACs/jumpTable.py
```diff
@@ -7,179 +7,201 @@
...
         self.whichDaisyBit = whichDaisyBit
         self.jumpIndex = nextJumpIndex
         self.bitOnOff = bool(bitOnOff)
-        
+
     def getWhichDaisyBit(self):
         return self._whichDaisyBit
+
     def setWhichDaisyBit(self, whichBit):
         if DAISY_VALUE_MIN < whichBit and whichBit < DAISY_VALUE_MAX:
```
maffoo: ditto

GHzDACs/jumpTable.py
```diff
@@ -189,48 +211,81 @@ class JumpTable(object):
...
+        """
+
+        :param startAddr: start address
+        :param jumps:
+        """
+        self._startAddr = 0
+        if counters is None:
+            self._counters = [0, 0, 0, 0]
+        else:
+            self._counters = list(counters)[0:4]
```
maffoo: `4` here should probably be a numeric constant, and technically it is board-dependent. Does that mean we need to have versioned jump-table implementations, like we do dac classes?

Qubit Server/dev/QubitServer.iml
```diff
@@ -0,0 +1,19 @@
...
+    <exclude-output />
+    <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$/src" isTestSource="false" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="module" module-name="JLabrad" />
+    <orderEntry type="library" name="guava-11.0.1" level="project" />
+  </component>
+</module>
```
maffoo: I'd suggest we don't put intellij files in repo

GHzDACs/Cleanup/dac.py
pomalley: PEP8: https://www.python.org/dev/peps/pep-0008/#inline-comments

(Also, yes.)

GHzDACs/Cleanup/dac.py
pomalley: Yes, but... the BIST is broken (it doesn't pass on the JT boards). I tried to figure out why and there's very little documentation out there for it, and even less of it is comprehensible, and the BIST code doesn't even follow that, AFAICT.

Qubit Server/dev/src/org/labrad/qubits/QubitContext.java
```diff
@@ -509,6 +500,38 @@ public void mem_sync_delay() {
...
+           doc = "Clear the jump table commands in this context.")
+  public void clear_jump_table() {
+    getExperiment().clearJumpTable();
+  }
+
+  @Setting(id = 371,
+           name = "Add Jump Table Entry",
+           doc = "Add a jump table entry to a single board.")
+  public void add_jump_table_entry(@Accepts({"s", "(ss)"}) Data channel,
+                                   @Accepts("s") Data command_name,
```
maffoo: This one can just be `String command_name`. Makes for less type-checking later (though you may want to check the value against some enum).

GHzDACs/Cleanup/dac.py
pomalley: pycharm complains when you don't call the superclass' constructor. But we don't want to do that here, because the JT stuff is different enough. Which maybe indicates that this isn't the proper class structure...

GHzDACs/Cleanup/dac.py
pomalley: I don't know, I need to talk to @DanielSank and @JulianSKelly about it. (Good thing I put the todo...?)

GHzDACs/ghz_fpga_server.py
pomalley: I don't exactly remember. I was trying to get the JT to work and... I don't remember. But it was necessary at the time. Probably.

maffoo:
qubit sequencer changes seem fine. Seems like backwards compatibility should be straightforward if you can just detect whether any jump table commands have been sent for a board, and if so send jump table commands to ghz_fpga_server, otherwise send mem and start delay as presently. We may have to get more sophisticated in the future when there are more board variations...

GHzDACs/jumpTable.py
pomalley: Yes, we probably should. There is some support for this in dac.py but I'm not sure the division of responsibility between dac.py and this (jumpTable.py) is entirely correct at the moment.

Qubit Server/dev/QubitServer.iml
pomalley: I think we should because it's really annoying to have to figure out (again and again) how to get the project to have the proper dependencies and spit out the jar file. Also we have eclipse's `.project` file in there...

maffoo: makes sense for now. Once we have a real build system that will actually generate project files and manage dependencies, this can go away (along with the eclipse stuff :) )

PR #40, opened 2015-02-16
Qubit server documentation

pomalley:
Here's a first shot at the qubit sequencer documentation.


maffoo:
LGTM. Thanks for adding these docs. I think there will be substantial changes to the eclipse/intellij sections as part of fixing #41, but having a discussion of the internals is very helpful.

PR #42, opened 2015-02-17
Build qubit server with sbt

### LINKED ISSUE #41: Set up build system for qubit sequencer
Right now, the qubit sequencer source lives here, and we also package up the server as an executable jar and commit the built jar to the repo as well. We should have a real build system that can build this in a reproducible way. I suggest we use [sbt](http://www.scala-sbt.org/) for this.

maffoo:
Move files in `Qubit Server/dev/src` to `Qubit Server/dev/src/main/java` as per the usual maven project organization conventions. Also add a simply sbt build description, including a reference to the jlabrad library published on bintray. This requires that we publish the jlabrad library itself to bintray; see https://github.com/martinisgroup/jlabrad/pull/2.

To build, you'll need to install sbt as described here: http://www.scala-sbt.org/0.13/tutorial/Setup.html
Then, cd into the `Qubit Server/dev` directory and run `sbt`. You can then run the `package` command to build the executable jar file (in `Qubit Server/dev/target`).

Fixes #41


pomalley:
LGTM pending the fact that there are a bunch of changes to the qubit sequencer in `jump_table_bench_test` and we'll have to deal with that rebase.

PR #44, opened 2015-02-17
Add GUIs to this repo.

pomalley:
I looked into keeping the git history of the files, but it would have included all the commit history for all of pyle, which I think would have swamped the servers history with a ton of unrelated commits.


pomalley:
OK let's do this because I've got the changes to the DR Logger / SuperMonitor depending on this.

joshmutus:
Just moving files into servers, LGTM

PR #56, opened 2015-03-10
Fix dr logger

pomalley:
I've tried to make the DR logger better about handling dead servers and displaying errors.

I've also changed the serial and mks servers to make the MKS server report 0 when the gauges read "LO" and the full scale when "OFF". Having NaNs in the data vault is really annoying. If there is a serial error, though, there still will be NaNs so you know if something is broken.

This includes commits from #44 so that needs to happen first.

### DIFF
--- a/GUIs/LabRADPlotWidget3.py
+++ b/GUIs/LabRADPlotWidget3.py
@@ -1,41 +1,59 @@
-import os, sys, time
+import sys
 
-import PyQt4.Qt as Qt
-import PyQt4.QtCore as QtCore
-
-from twisted.internet.defer import inlineCallbacks, returnValue
-
-import numpy as np, matplotlib as mp, matplotlib.pyplot as plt
+from PyQt4 import Qt, QtCore
+from twisted.internet.defer import inlineCallbacks
+import numpy as np
+import matplotlib.pyplot as plt
 from matplotlib.backends.backend_qt4agg import FigureCanvasQTAgg as FigureCanvas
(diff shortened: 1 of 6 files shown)


ejeffrey:
I don't like reporting full scale.  This causes the plotters to rescale
badly.  NaN is the correct result when the data is unavailable, what is the
problem with having it in the datavault?

pomalley:
Just that the grapher can't handle it (but the supermonitor should be able
to just drop the NaNs). The grapher is the easiest way to look through
historical cooldowns. (But I have a PR for some plotting tools to help with
this, iirc.)

pomalley:
But since both you and dan don't like it maybe we should revert that change.

ejeffrey:
I think we should just fix supermonitor to support easy plotting at an
arbitrary time interval, then we don't have to worry about grapher for this
task.  Putting wrong data in the datavault seems like the wrong solution.
In addition to breaking auto-scaling plots, it will silently break
analysis/fitting.  Using NaNs will either work (some fitting libraries skip
NaNs because they are the standard way to represent missing data), or cause
errors if you don't manually filter them.  This is definitely better than
producing e.g., a fit curve that is hugely off because one data point is
10^308.

pomalley:
I'm ok with that, but to be clear, when the gauge reports the reading as
off scale ("OFF") it reports the gauge's full scale (which is queried from
the gauge itself), not 10^308 or whatever. So in terms of graphing it
should look fine, just like you've railed the sensor.

PR #65, opened 2015-03-19
Add averaging sram border functionality.

JulianSKelly:
When the DACs reach the end of SRAM, they idle by playing back the last 4 values repeatedly. If these 4 values are different, you get 250 MHz noise. Previously, we handled this by setting the border values to zero, however now I want to be able to put out a DC offset when the SRAM is done. So I've added functionality to average these values which prevents 250 MHz noise.


DanielSank:
These kinds of changes need documentation.

JulianSKelly:
Changed borderValues to dict, with 'start' and 'end' keys. Hopefully this is more clear.

maffoo:
It seems to me that this setting should not be set at the server or board level, but rather something one would want to control per sequence. Also, "border" in the very near future is going to mean something different from what it does now, because with the jump table many SRAM segments will be internal, followed by other segments, and only the last segment output before the board starts idling would want this kind of averaging treatment at the end.

JulianSKelly:
With these concerns in mind, can I get assistance to make this better?

Thanks.

JulianSKelly:
This broke dual block. Working on fixing it.

pomalley:
> This broke dual block.

Where have I heard this before? ;-)

JulianSKelly:
@pomalley errday

DanielSank:
^ chudat.

pomalley:
"Dual block: turning every day into errday since 1897."

DanielSank:
What happened in 1897?

pomalley:
Dunno, it was the first year that came into my head. Actually 2008? 2006?

maffoo:
this change LGTM. The only thing to consider is that by removing the borderValues setting, old versions of pyle will no longer work with the server, so we have an issue of sequencing the server deployment and people updating their pyle code. I might suggest leaving the borderValues setting and just having it ignore the passed in values; that won't actually change the sequences in a meaningful way but will allow slightly older code to continue working. Then, file a separate issue to actually remove the setting after people have been given a chance to upgrade all their code to the latest version.

DanielSank:
While I normally really don't like silent "errors" like this, I think @maffoo's strategy is the only one that can really work.

JulianSKelly:
Good idea. Added borderValues call back in, and it does nothing but print that it is deprecated.

PR #66, opened 2015-03-20
Add missing exception in sequenceTime.

JulianSKelly:


GHzDACs/Cleanup/dac.py
```diff
@@ -1064,16 +1064,18 @@ def cmdTime_cycles(cmd):
...
+        elif opcode == 0x3:
             return MemorySequence.getAddress(cmd) + 1
         #run sram
-        if opcode == 0xC:
+        elif opcode == 0xC:
             # TODO: Incorporate SRAMoffset when calculating sequence time.
             #       This gives a max of up to 12 + 255 us
             return 25*12 # maximum SRAM length is 12us, with 25 cycles per us
+        else:
+            raise Exception("Unkown opcode")
```
maffoo: "Unknown"

Also, I would suggest including the bad opcode here:

```python
raise Exception("Unknown opcode: {}".format(opcode))
```

Otherwise, LGTM

JulianSKelly:
@maffoo fixed

maffoo:
still a typo: "Unkown" should be "Unknown" :)

also, I think we should get in the habit of using `format` instead of `%` for string formatting (see https://www.python.org/dev/peps/pep-3101/). At some point we can go through and fix old code, but for now, I'd suggest at least using `format` on new code.

these are small issues, though. consider this an LGTM

PR #71, opened 2015-03-24
Fixing units in agilent_DSO7104B

jwenner:
Fixing units in agilent_DSO7104B/position and get_trace so that compatible with latest pylabrad.

Note that position (according to documentation), which modifies where y=0, is in units of divisions (hence why everything is made dimensionless).

For get_trace, output is expected to have units of mV and ns, which is now satisfied with this change.

-Jim/Charles


DanielSank:
How about adding support for this scope to the new shiny oscilloscope server which is supposed to support all scopes? You can find the WIP in the obviously named branch :)

maffoo:
LGTM

also, +1 for supporting scopes in a single common server

DanielSank:
@maffoo Yeah I'm excited about that. I haven't spent as much time as I wish I could. The code is there and works for at least the scope on Ivan. Adding new scope support is easy due to the modular design. You just make a new DeviceWrapper subclass and write a few GPIB strings.

jwenner:
@maffoo, does this mean I can now merge in? If so, I'm assuming I should change the commit message title to "Fix units...", note the file in the title, and note your review at the bottom. Anything else? Thanks.

PS: I agree that the common server would be nice. But @DanielSank, for someone who's job is getting out of here...

DanielSank:
> PS: I agree that the common server would be nice. But @DanielSank, for someone who's job is getting out of here...

Good point :-)

maffoo:
yeah, please feel free to merge.

PR #73, opened 2015-03-25
Add missing setSettling call in dacify

JulianSKelly:
We never saw this before as we always used dacify_ft. We now use the time domain version.


pomalley:
LGTM

JulianSKelly:
done via command line.

PR #76, opened 2015-03-27
Fix jqs_test script to run code inside a main method.

maffoo:
Otherwise the script gets discovered by pytest and it's not happy...


pomalley:
LGTM

PR #82, opened 2015-04-01
Add more descriptive error message when unknown fpga builds are detected.

### LINKED ISSUE #81: Better FPGA server error message for unknown build number
If the FPGA server finds a DAC with a build number it doesn't have a class for, you get the error message: "There was an error: ('DAC', 15)" (if 15 is the build number). This should be made more explicit.

maffoo:
Fixes #81


pomalley:
LGTM--this is basically what I did in the JT branch, but better--but I wonder about just merging in changes to such critical servers without testing. Anyone have ideas about this? (Of course, "make real tests" would be ideal, but...)

maffoo:
I'm with you on better testing, but I think in this case the change is pretty low-risk. If it breaks then we just get a bad error message, which is at least no worse than the problem we're trying to fix :)

pomalley:
Yeah, I'm OK with this.

PR #85, opened 2015-04-06
Fixed a bit rounding issue in sram edges, added support for reading a dither=True/False reg key.

rb25:


JulianSKelly:
LGTM.

ghzdac/correction.py
```diff
@@ -782,16 +780,16 @@ def DACifyFT(self, carrierFreq, signal, t0=0, n=8192, loop=False,
                 self.min_rescale_factor = rescale
             fullscale *= rescale
 
-
+        #due to deconvolution, the signal to put in the dacs can be nonzero at the end of a sequence with even a short pulse. 
+        #This nonzero value persists, even when running the board with an empty envelope. To remove this, the first and last 4 (FOUR) values must be set.
+        i[:4] = 0.0
+        i[-4:] = 0.0
+        q[:4] = 0.0
+        q[-4:] = 0.0
```
maffoo: why set these to 0 instead of the calibrated zero values?

JulianSKelly: The lines 789-790 add the zero values back in, and get the rounding.

DanielSank:
I'm really nervous about people making changes to mission critical code and getting LGTM's from their direct work partners I think we need a little more separation there. Do others agree?

maffoo:
I would tend to agree in general; a pair (or two) of "fresh" eyes is a good idea when changing code that everyone relies on. I think the change here is fine as it doesn't change the default dithering behavior, just makes it configurable.

DanielSank:
Right, I'm not worried about this particular change. I just get nervous when I see changes to the deconvolution code submitted and approved by the same team. This tends to push the code into situations where it serves one project's specific goals at the cost of others.

PR #86, opened 2015-04-08
Fix some imports broken by pyle registry refactor.

maffoo:
These pyle imports were broken by recent refactoring in pyle, but also seem to be unnecessary. There's one remaining import in MBEStatus that is still used: datavault. It's unfortunate to have this dependency from servers on pyle. Maybe the datavault wrapper stuff should eventually be moved here to live with the data vault server?

### DIFF
--- a/GRAPE/GRAPE.py
+++ b/GRAPE/GRAPE.py
@@ -37,6 +37,8 @@
 # > grape.set_control_z_parameters(Value(40.0,'ns'), ..., ..., ...)
 # > result = grape.controlz(index1, index2)
 
+import os
+
 import numpy as np
 
 from labrad.units import Value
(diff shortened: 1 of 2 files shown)


DanielSank:
+1 for moving the datavault tools into the same place as the datavault server code.

DanielSank:
LGTM, but not sure how to test.

ejeffrey:
The servers directory serves a weird purpose, in that the node scans it for
servers it can start.  I am all for moving client utility code out of pyle,
but I don't think it should be mixed in with the servers, although it could
be in the same repo, if we renamed the repo labrad-utils, then made servers
a subdirectory of that.

DanielSank:
Hey, that's a good idea, although I'd go for something like `labrad-modules`.

maffoo:
Opened #109 to discuss reorg or the servers repo. This can also include renaming it to better reflect its purpose.

PR #88, opened 2015-04-11
Make use of microwave switch optional in calibration

DanielSank:


ghzdac_cal/GHz_DAC_calibrate.py
```diff
@@ -26,9 +33,9 @@
 # GHz_DAC_calibrate is only a front-end, the actual calibration
 # routines are in ghzdac.calibrate
 
-import ghzdac
-import ghzdac.calibrate as calibrate
-import ghzdac.keys as keys
+import servers.ghzdac as ghzdac
+import servers.ghzdac.calibrate as calibrate
+import servers.ghzdac.keys as keys
```
maffoo: Are these imports correct? This assumes that the servers repo is cloned as `servers` and the parent directory is on the python path. If we want to actually put the code in this repo in a package, then I think we should reorganize this repo to match the structure in the `pylabrad` and `pyle` repos where there is a package *directory* in the repo with an `__init__.py` file, rather than putting the `__init__.py` at the repo top level, as here.

maffoo:
Other than the imports, this LGTM. I would be ok with committing this, but I think we need to have a larger discussion about PYTHONPATH organization, so we can have a common setup, get imports right, and have a sane separation between runnable scripts and library code.

ghzdac_cal/GHz_DAC_calibrate.py
DanielSank: I agree that this repo should be reorganized. However, given the current format (and reasonably common practice), I do indeed expect the repo to be inside a directory (i.e. `src`) which is in the python `sys.path`.

Again, I agree that this ought to be fixed.

DanielSank:
> Other than the imports, this LGTM. I would be ok with committing this, but I think we need to have a larger discussion about PYTHONPATH organization, so we can have a common setup, get imports right, and have a sane separation between runnable scripts and library code.

+1.2385

PR #89, opened 2015-04-11
Ensure Hittite frequency write is integer number in Hz

DanielSank:
There is a deeper issue here: servers which set hardware parameters should *always* read back the parameter they try to write and **fail** if the attempted write value does not match the read-back.

I think the change proposed here is better than what was there before because in the old way the result was that the write would silently do nothing. Now you at least get a frequency which is within 1 Hz of what you wanted. I'm all ears if folks think this sucks. If we had a LabRAD type for unit-checked integers this would be much easier to deal with, but we don't. Ideas?

### DIFF
--- a/HittiteT2100/hittite_T2100.py
+++ b/HittiteT2100/hittite_T2100.py
@@ -65,9 +65,9 @@ class HittiteWrapper(GPIBDeviceWrapper):
         
     @inlineCallbacks
     def setFrequency(self, f):
-        f = f['Hz']
+        f = int(f['Hz'])
         if self.frequency != f:
-            yield self.write('SOUR:FREQ:FIX %f' % f)
+            yield self.write('SOUR:FREQ:FIX %d' % f)
             self.frequency = f
     
     @inlineCallbacks


ejeffrey:
I disagree.  I don't think it is generally useful or possible to insist
that the floating point representation match an exact digital value.  If
rounding a value to the nearest representation doesn't make sense, the
server setting should be changed to take an integer.  Any code which
expects floating point values to have exact results is broken.

Consider this: the fastbias cards cannot be set to 1.0 volts.  The nearest
legal value is approximately 999.985 mV.  Should that server really fail if
you ask it to produce 1 volt?

There are basically three ways to deal with this:

1) write code whose correctness does not depend on small changes needed to
coerce to a legal value
2) Eliminate floating point numbers and used only fixed point for anything
that will be used to set a hardware value
3) Query the device after setting it to learn the actual value it was
configured for, and base further computation off of that.  Keep in mind
that devices like frequency generators that use an internal decimal
representation may not return something that is exactly representable as a
floating point number, so this may not even fix the problem.

Can you file an issue on the actual problem that lead to this change?

Evan

DanielSank:
@ejeffrey see #89 .

DanielSank:
Will discuss in issue #90 and make a new PR when we're ready.

PR #96, opened 2015-04-14
Remove a bunch of old and unused servers

maffoo:
The old python qubit server and several other servers that use it have long since become obsolete. I also removed some delphi binary servers which are not needed, and which don't really belong in a source repo anyway. And finally, a couple of other servers that AFAICT are not being used and/or are just test or example code; since there are some example servers bundled with pylabrad itself, these seem unnecessary.


ejeffrey:
LGTM.

jwenner:
LGTM.

Does anyone (e.g., @maffoo, @DanielSank) actually know about the IBCL GPIB servers (as opposed to the regular GPIB servers)? Can we get rid of these as well?

DanielSank:
LGTM

ejeffrey:
@jwenner

It looks like IBCL is an old serial-to-GPIB interface made by NI.  The IBCL
server accesses it via the serial port rather than through VISA layer, I
don't know if VISA supports it or not.  The manual I found was copyright
1993 (!).

In general I don't see a compelling reason to remove servers for hardware
we no longer use, but it seems likely that nobody uses this (the list of
serial ports is hard coded into the server...).

Evan

PR #101, opened 2015-04-17
GHzFPGA.run: Enable ADC Average Mode

### LINKED ISSUE #100: ADC Average Mode Broken
When try to use fpga average mode, get error in ghz_fpga_server since demod channel not specified. On branch 00d5ad5e77ec5f6dc8030fa0798d063b064bdc4f.
```
In [25]: fpgaTest.average(s,cxn,3,True)
---------------------------------------------------------------------------
Error                                     Traceback (most recent call last)
<ipython-input-25-53c37bbfb41c> in <module>()
----> 1 fpgaTest.average(s,cxn,3,True)

C:\pyle\pyle\scripts\fpgaTest\fpgaTest.py in average(sample, cxn, reps, plot, na
me, dacsAdcs)
   1086     fpga.daisy_chain(daisyChainList)
   1087     fpga.timing_order([adc['_id']])
-> 1088     data = fpga.run_sequence(reps, True)[0]
   1089     if plot:
   1090         plt.figure()

C:\pylabrad\labrad\client.pyc in __call__(self, *args, **kw)
     62         if wrap:
     63             future.addCallback(lambda resp: resp[0][1])
---> 64         return future.wait() if wait else future
     65
     66     # data to be loaded on demand

C:\pylabrad\labrad\backend.pyc in wait(self)
    397                 # If any Future in the queue fails,
    398                 # we immediately bail.
--> 399                 result.raiseException()
    400             else:
    401                 for func, args, kw in f.callbacks:

C:\Python27\lib\site-packages\twisted\python\failure.pyc in raiseException(self)


Error: (0) [GHz FPGAs] Remote Traceback (most recent call last):
  File "C:\pylabrad\labrad\server.py", line 262, in handleRequest
    result = yield setting.handleRequest(self, c.data, data)
  File "C:\Python27\lib\site-packages\twisted\internet\defer.py", line 1105, in
_inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "C:\Python27\lib\site-packages\twisted\python\failure.py", line 389, in t
hrowExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "C:\pylabrad\servers\GHzDACs\ghz_fpga_server.py", line 1449, in run_seque
nce
    timingOrder)
  File "C:\Python27\lib\site-packages\twisted\internet\defer.py", line 1107, in
_inlineCallbacks
    result = g.send(result)
  File "C:\pylabrad\servers\GHzDACs\ghz_fpga_server.py", line 758, in run
    raise RuntimeError('channel format not understood')
exceptions.RuntimeError: channel format not understood
 [payload=None]

In [26]:
```

jwenner:
Fixes #100.

In FPGA server/run (and hence run_sequence), if ADC in average mode so that no demod channel specified, raised 'channel format not understood' exception. Now add elif case in getTimingData so if ADC is present in channel name but not :: (for demodulation), can still return data. Move calls to channel to if statement for if channel defined. Amend return format type for run_sequence to also allow '*3i' for average mode data.

Note that in going between demod and average modes (although in different scans & calls to run_sequence), it seems to work fine (although it's possible I might not be seeing/knowing what @JulianSKelly was referring to).

### DIFF
--- a/GHzDACs/ghz_fpga_server.py
+++ b/GHzDACs/ghz_fpga_server.py
@@ -740,55 +740,46 @@ class BoardGroup(object):
             results = yield readAll # wait for read to complete
             
             if getTimingData:
-                # print("fpga server: run: %s"%(timingOrder,))
                 answers = []
                 # Cache of already-parsed data from a particular board.
                 # Prevents un-flattening a packet more than once.
                 extractedData = {}
                 for dataChannelName in timingOrder:
-                    # If dataChannelName has a :: in it, it's an ADC with specified demod
-                    # channel
                     if "::" in dataChannelName:
+                        # If dataChannelName has :: in it, it's an ADC
+                        # with specified demod channel
                         boardName, channel = dataChannelName.split('::')
                         channel = int(channel)
                     elif 'DAC' in dataChannelName:
                         raise RuntimeError("DAC data readback not supported")
                     elif 'ADC' in dataChannelName:
+                        # ADC average mode
                         boardName = dataChannelName
                         channel = None
                     else:
                         raise RuntimeError('channel format not understood')
                     
-                    # If we have already parsed the packet for this board, fetch the
-                    # cached result.
                     if boardName in extractedData:
+                        # If we have already parsed the packet for this
+                        # board, fetch the cached result.
                         extracted = extractedData[boardName]
-                    # Otherwise, extract the data, cache it, and add the relevant part
-                    # to the list of returned data.
                     else:
+                        # Otherwise, extract data, cache it, and add
+                        # relevant part to the list of returned data
                         idx = boardOrder.index(boardName)
                         runner = runners[idx]
                         result = [data for src, dest, eth, data in results[idx]['read']]
-                        # print("fpga server: run: result is %s"%(result,))
                         # Array of all timing results (DAC)
                         extracted = runner.extract(result)
                         extractedData[boardName] = extracted
-                        
-                        # XXX I'm about to re-define extracted. This is ok
-                        # because it's a tuple and won't mutate the thing in the
-                        # cache, but it's bad and I should rename the variables.
-                        
-                        # XXX Ranges code needs to be updated. Next line obsolete.
-                        # runner.ranges = extracted[1]
-                        # Add extracted data to the list of data to be returned.
-                        # If this is an ADC demod channel, grab that channel's
-                        # data only
-                        # print("fpga_server: extracted: %s"%(extracted,))
-                        # print("fpga_server: extacted no pkt counters: %s"%(type(extracted),))
-                        # print("fpga_server: channel: %d"%(channel,))
+                    # Add extracted data to list of data to be returned
                     if channel != None:
-                        extracted = extracted[0][channel]
-                    answers.append(extracted)
+                        # If this is an ADC demod channel, grab that
+                        # channel's data only
+                        extractedChannel = extracted[0][channel]
+                    else:
+                        extractedChannel = extracted
+                    answers.append(extractedChannel)
                 returnValue(tuple(answers))
         finally:
             self.pipeSemaphore.release()
--- a/GHzDACs/ghz_fpga_server.py
+++ b/GHzDACs/ghz_fpga_server.py
@@ -1361,34 +1361,39 @@ class FPGAServer(DeviceServer):
                      setupState=[]):
         """Executes a sequence on one or more boards.
 
-        reps:
-            specifies the number of repetitions ('stats') to perform
-            (rounded up to the nearest multiple of 30).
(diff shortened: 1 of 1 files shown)


DanielSank:
Why is the return type `*3i` for average mode? Shouldn't it be `*2i`?

jwenner:
@DanielSank, I agree it should be ```*2i```. However, at least fpgaTest assumes that it be ```*3i```, so I would propose that we stick with this for backwards compatibility.

Actually, before average mode was removed, the type tag for average mode was ```?```.

DanielSank:
> @DanielSank, I agree it should be *2i. However, at least fpgaTest assumes that it be *3i, so I would propose that we stick with this for backwards compatibility.

I think we should fix it. What's the extra dimension for?

jwenner:
@DanielSank, the dimensions from run_sequence appear to be as follows:
   * demod: [demodChannel, trialNumber, data?, data?]
   * average: [demodChannel, IOrQ, data]

I would propose that we instead make the average mode data 4-dimensional of the same form as demod (i.e., [demodChannel, trialNumber, IOrQ, data]), with the length=1 in the demodChannel and trialNumber dimensions. The reason would be so that we could (maybe?) take average mode & demod mode data at the same time on different ADCs (e.g., measure qubit state &, say, resonator reflection at the same time).

Would changing any of this (in any direction) require changes in the qubit sequencer server?

Thanks.

DanielSank:
> I would propose that we instead make the average mode data 4-dimensional of the same form as demod (i.e., [demodChannel, trialNumber, IOrQ, data])

The demodulator mode data indices are `(channel, stat, demod, I/Q)`. I'm not sure what you mean by `data` for the final index. I also don't understand what it means to have `channel` and `trialNumber` indices on average mode data, as the ADC board does not support separate traces for different trials. In other words, average mode doesn't have "stats" in the same was as demodulator mode.

jwenner:
@DanielSank, happy now?

PS: Upon thinking about it, I believe the third dimension was for if we had multiple ADCs present, from which we were reading back average mode data simultaneously.

ejeffrey:
As far as I know, we can't currently support having some boards in average mode and some in IQ mode without major changes, and even if we did, we couldn't return the data in a 4D array since the channels would not all be the same length.  I don't think trying to be future compatible in this way is going to help us any.

Having a third index for 'board number' in the case of multiple boards is fine, but in any case the axes of the return data should be documented!  If it said "the return in average mode is a 3D array of (board#, time, I/Q) data" we wouldn't be having this conversation. Same for the IQ mode.

jwenner:
@DanielSank, @ejeffrey, could you please now take a look at this ghz_fpga_server update?

* Gone back to the 3D array to handle the multiple ADC boards case. 
* Updated the run_sequence docstring describing the return data format so it's no longer completely out of date. 
* Cleaned up some of the comments in the part of run I've dealt with (including renaming one variable which was multiply defined w/ a comment saying why OK but bad).

jwenner:
@DanielSank, if/when you approve this and I merge it back in, can we also close martinisgroup/pyle#103 as well (considering the updates to the fpga.run_sequence docstring)? Thanks.

DanielSank:
@jwenner This looks great. Thanks for moving the comments around so that they make more sense.

I would like to fix up the documentation of the `Run Sequence` setting, as the notation used there for describing the data format seems to be in Jim Wenner Markdown<sup>TM</sup>. If you don't mind I'll add a commit to this branch.

DanielSank:
LGTM.

Jim, when rebasing please try to keep the changes in content separate from the documentation and white space.

This is great, thanks.

zchen088:
>Jim, when rebasing please try to keep the changes in content separate from the documentation and white space.

This is a thing we're supposed to do when rebasing? I thought the whitespace thing was only for the initial reviewing, and I've been squashing all of my pull requests down to one commit. I also wasn't aware that documentation needs to be in a separate commit. Can this get clarified in the contributing doc?

Also, can we get an explanation in the contributing guide as to why this needs to be the case?

jwenner:
@DanielSank, most of your commit looks good. However, there was an error in your revision to the docstring for average mode (order of indices) that I've now fixed, and I also added a sentence to clarify what the retrigger index is. Can you take a quick look and tell me if this is OK?

For the documentation commit, can I mark martinisgroup/pyle#103 closed?

Thanks.

DanielSank:
> However, there was an error in your revision to the docstring for average mode (order of indices) that I've now fixe

@jwenner Thanks for fixing that mistake.

> For the documentation commit, can I mark martinisgroup/pyle#103 closed?

Yes, as soon as we merge.

> Can you take a quick look and tell me if this is OK?

Yeah, looks great.

Issue #98, opened 2015-04-16
dr_logger.py is very hard to understand


DanielSank:
We're having trouble getting the supermonitor to run and I think it's because the dr_logger is crashing ungracefully when it can't access some resources. I am trying to fix this but having a hard time following the code.

@pomalley can you spare 30 minutes tomorrow to extreme-coding this thing into shape?

pomalley:
Grab me at any point today.

pomalley:
I'd argue that it's only marginally more understandable more... ;)

DanielSank:
Yeah, I didn't mean to close this. Accident.

PR #102, opened 2015-04-17
Fix dr logger

DanielSank:
Addresses #98 to some extent. More documentation still needed, but at least this regularizes the expected registry format.

NB: This commit changes the expected form of the registry entries for the dr logger (in a good way). See server docstring for expected registry format.

@pomalley


DanielSank:
@pomalley bump

PR #104, opened 2015-04-19
Make averaging, dithering, and zeroing optional and off by default.

DanielSank:
The deconvolution server was mutating the user's waveform data without respect to any parameters passed into `Correct` or `Correct FT`:

* For Z waveforms, dithering had the appearance of being set by a boolean parameter on the context, but was actually being defaulted to `True` at context initialization! It was also optionally set by a global registry key for the entire server. This should be a per-request variable.
* For Z waveforms, averaging the first and last four SRAM samples was done without regard for any input parameters. This meant that there was *no way* to not have your pulse sequence mutated by the server.
* For IQ waveforms, zeroing the first and last four SRAM samples was done without regard for any input parameters. This meant that there was *no way* to not have your pulse sequence mutated by the server.

This is problematic: anyone trying to use the deconvolution server outside of qubit experiments had their waveforms mysteriously mutated. Note that this is *not* just me. Other people are starting to use fpgaTest for bringup and debugging purposes. We need to be mindful of everyone's use cases when modifying our code base. Aside from that, it is not good software design to bake unexpected behavior into the code.

This change makes all of these mutations optional and off by default. This means that the qubit sequencer code needs to be updated to turn these features on.

@pomalley, would you mind doing that on this branch?

@maffoo @ejeffrey @JulianSKelly @rb25 I will not merge these changes until all of you sign off on it. This is an important and complex piece of code. In fact, I propose that we set this (and in the future other) pieces of code such that they are immutable without LGTM from at least three of the people pinged here.

### DIFF
--- a/Qubit Server/dev/src/main/java/org/labrad/qubits/QubitContext.java
+++ b/Qubit Server/dev/src/main/java/org/labrad/qubits/QubitContext.java
@@ -559,9 +559,12 @@ public class QubitContext extends AbstractServerContext {
           + "microwave quadratures.  The length of the data should "
           + "match the length of the current SRAM block.  "
           + "An optional boolean specifies whether the data "
-          + "should be deconvolved (default: true)."
-          + "If deconvolve=false, the data can be specified as DAC-ready"
-          + "I and Q integers.")
+          + "should be deconvolved (default: true).  "
+          + "If deconvolve=false, the data can be specified as DAC-ready "
+          + "I and Q integers. "
+          + "If zeroEnds is true (the default), then the first and last "
+          + "4 nanoseconds of the deconvolved sequence will be set to the "
+          + "deconvolved zero value, to ensure microwaves are turned off.")
   public void sram_iq_data(
       @Accepts({"s", "ss"}) Data id,
       @Accepts("*c") Data vals
@@ -605,7 +608,10 @@ public class QubitContext extends AbstractServerContext {
           + "The data is specified as a list of complex numbers, "
           + "with the real and imaginary parts giving the I and Q "
           + "microwave quadratures.  The length of the data should "
-          + "match the length of the current SRAM block.")
+          + "match the length of the current SRAM block.  "
+          + "If zeroEnds is true (the default), then the first and last "
+          + "4 nanoseconds of the deconvolved sequence will be set to the "
+          + "deconvolved zero value, to ensure microwaves are turned off.")
   public void sram_iq_data_fourier(
       @Accepts({"s", "ss"}) Data id,
       @Accepts("*c") Data vals,
@@ -636,9 +642,13 @@ public class QubitContext extends AbstractServerContext {
           + "\n\n"
           + "The length of the data should match the length of the "
           + "current SRAM block.  An optional boolean specifies "
-          + "whether the data should be deconvolved (default: true)."
+          + "whether the data should be deconvolved (default: true).  "
           + "If deconvolve=false, the data can be supplied as DAC-ready "
-          + "integers.")
+          + "integers.  "
+          + "If averageEnds is true (the default), then the first and last "
+          + "4 nanoseconds of the deconvolved sequence will be averaged and "
+          + "set to the same value, to ensure the DAC outputs a constant "
+          + "after the sequence is run.")
   public void sram_analog_data(
       @Accepts({"s", "ss"}) Data id,
       @Accepts("*v") Data vals
@@ -682,7 +692,11 @@ public class QubitContext extends AbstractServerContext {
           + "\n\n"
           + "Because this represents real data, we only need half as many samples.  "
           + "In particular, for a sequence of length n, the fourier data given "
-          + "here must have length n/2+1 (n even) or (n+1)/2 (n odd).")
+          + "here must have length n/2+1 (n even) or (n+1)/2 (n odd)."
+          + "If averageEnds is true (the default), then the first and last "
+          + "4 nanoseconds of the deconvolved sequence will be averaged and "
+          + "set to the same value, to ensure the DAC outputs a constant "
+          + "after the sequence is run.")
   public void sram_analog_fourier_data(
       @Accepts({"s", "ss"}) Data id,
       @Accepts("*c") Data vals,
--- a/Qubit Server/dev/src/main/java/org/labrad/qubits/QubitContext.java
+++ b/Qubit Server/dev/src/main/java/org/labrad/qubits/QubitContext.java
@@ -562,22 +562,34 @@ public class QubitContext extends AbstractServerContext {
           + "should be deconvolved (default: true)."
           + "If deconvolve=false, the data can be specified as DAC-ready"
           + "I and Q integers.")
-  public void sram_iq_data(@Accepts({"s", "ss"}) Data id,
-      @Accepts("*c") Data vals) {
+  public void sram_iq_data(
+      @Accepts({"s", "ss"}) Data id,
+      @Accepts("*c") Data vals
+  ) {
     sram_iq_data(id, vals, true);
-    sramDirty = true;
   }
   @SettingOverload
-  public void sram_iq_data(@Accepts({"s", "ss"}) Data id,
+  public void sram_iq_data(
+      @Accepts({"s", "ss"}) Data id,
       @Accepts({"*c", "(*i{I}, *i{Q})"}) Data vals,
-      boolean deconvolve) {
+      boolean deconvolve
+  ) {
+    sram_iq_data(id, vals, deconvolve, true);
+  }
+  @SettingOverload
+  public void sram_iq_data(
+      @Accepts({"s", "ss"}) Data id,
+      @Accepts({"*c", "(*i{I}, *i{Q})"}) Data vals,
+      boolean deconvolve,
+      boolean zeroEnds
+  ) {
     IqChannel ch = getChannel(id, IqChannel.class);
     if (vals.isCluster()) {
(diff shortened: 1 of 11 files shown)


DanielSank:
Fixes #107.

DanielSank:
I did test this on Jules, but not with the qubit stack. For the qubit stack to work, we need to have fpgaseqtransmon pass a few extra parameters to the qubit sequener, and from there to the deconvolution server.

dac_calibration_server.py
```diff
@@ -86,15 +86,12 @@ def initServer(self):
         self.IQcalsets = {}
         self.DACcalsets = {}
         print 'loading server settings...',
-        yield self.loadServerSettings()
+        self.loadServerSettings()
```
maffoo: why are you not yielding? `self.loadServerSettings` returns a deferred, so unless you yield errors will go into the ether.

maffoo:
I think we need to be able to control averaging and zeroing separately for the two ends of a sequence to be deconvolved. This will certainly be needed for some jump table operations where a given segment may not start and end at the same levels. Also, given that zeroing and averaging are mutually exclusive, it would seem to me better to specify a boundary treatment for each end, something like:

```python
boundary_treatment = ('zero', 'average')
```

So, `boundary_treatment` would be of type `(ss)` where each string could be one of `'none'`, `'zero'` or `'average'`. It would default to `('none', 'none')`.

DanielSank:
@maffoo I'm on gchat. Let's talk about this.

rb25:
I think that the average mode should be removed, and that the ends should be zero'ed by default.

Suppose you run an experiment, and because of deconvolution, the ends are different: as a result, a 250 MHz tone, among others will be coming out of the DAC. This is undesirable for both Z and XY boards, so something needs to be done **by default**

Ok, say we average the last 4 values. Because of deconv, the last 4 values probably have an average which isn't zero. 
-For Z boards: the qubit frequency is shifted slightly, so if, the next time, you run only XY boards (because you're done scanning w Z), you have the wrong qubit frequency.
-For XY boards, you're off from the optimal value to suppress carrier leakage. So you get carrier leakage, and probably the next pi pulse isn't optimal too

So: we need to do something about it and not average. That leaves zeroing at the end. We need to do that by default, because the only time you don't need it is when using the boards for CW pumping of paramps (which we're probably going to do pulsed anyway). I agree that there needs to be a toggle in the registry to turn it off.

About dithering: we've experimentally found that dithering with an amplitude 1 to 1.5 bits for Z works better than w/o dithering. I think we want to enable that by default.

DanielSank:
Just had a talk with Rami. I don't think he understood that modifying how the qubit sequencer invokes the deconvolution code is part of the proposed set of changes here. Given our exchange just now I think we can move on with the original plan to have mutations off by default.

DanielSank:
I don't like the name `Correct Z` for single channel data. The term "Z" only makes sense if you're controlling a qubit. We (and other groups!) use the GHz DACs to make control pulses for all kinds of stuff. Anyone have an idea for a better name?

ejeffrey:
Some places in the code (the qubit sequencer?) call them 'analog'
channels.  It isn't the best name (IQ channels are "analog" too), but I
think if you see "analog" next to "IQ" it will be clear what is going on,
and if you think about it, digital-to-analog is the basic functionality,
and IQ is something layers on top of that.

maffoo:
'Correct Analog' works for me.

DanielSank:
Yeah, I like that. Will do.

dac_calibration_server.py
```diff
@@ -102,15 +99,11 @@ def loadServerSettings(self):
...
         }
         for key in keys.SERVERSETTINGVALUES:
             default = defaults.get(key, None)
-            keyval = yield reg.get(key, False, default)
-            if not isinstance(keyval, bool):
-                #keyval is a number, in labrad units
-                keyval = keyval
+            keyval = default
             print key, ':', keyval
             d[key] = keyval
```
maffoo: I just realized that this whole thing is kinda bogus; why not get rid of the `keys.SERVERSETTINGVALUES` array and build the `serverSettings` dict directly:
```python
self.serverSettings = {
    'deconvIQ': True,
    ...
    'dither': False
}
```

DanielSank: Yes, it's bogus. I didn't refactor it yet because I am trying to keep the commits very small to facilitate reviews.

dac_calibration_server.py
```diff
@@ -259,11 +251,42 @@ def dac(self, c, dac):
         c['DAC'] = dac
         return dac
 
+    @setting(30,
+        'Correct IQ',
+        data=['*(v, v): I/Q data', '*c: I/Q data'],
+        zero_ends='b',
+        returns=['(*i, *i): Dual channel DAC values'])
```
maffoo: This is so much nicer to have a single well-defined return value. Makes it conceivable to move to a more static type system or IDL...

DanielSank:
@ejeffrey @JulianSKelly please look at this soon.

dac_calibration_server.py
DanielSank: Yes.

Qubit Server/dev/src/main/java/org/labrad/qubits/QubitContext.java
```diff
@@ -562,22 +562,34 @@ public void sram_dual_block_delay(@Accepts("v[ns]") double delay) {
...
           + "If deconvolve=false, the data can be specified as DAC-ready"
           + "I and Q integers.")
-  public void sram_iq_data(@Accepts({"s", "ss"}) Data id,
-      @Accepts("*c") Data vals) {
+  public void sram_iq_data(
+      @Accepts({"s", "ss"}) Data id,
+      @Accepts("*c") Data vals
+  ) {
     sram_iq_data(id, vals, true);
-    sramDirty = true;
```
pomalley: Why not set sramDirty?

Qubit Server/dev/src/main/java/org/labrad/qubits/QubitContext.java
```diff
@@ -594,12 +606,23 @@ public void sram_iq_data(@Accepts({"s", "ss"}) Data id,
           + "with the real and imaginary parts giving the I and Q "
           + "microwave quadratures.  The length of the data should "
           + "match the length of the current SRAM block.")
```
pomalley: We should update the doc to indicate zeroEnds, and especially the default.

Qubit Server/dev/src/main/java/org/labrad/qubits/QubitContext.java
maffoo: it gets set by the method that we called on the previous line. Incidently, all this boilerplate goes in scala because it has optiona parameters with default values.

Qubit Server/dev/src/main/java/org/labrad/qubits/QubitContext.java
maffoo: will do

Qubit Server/dev/src/main/java/org/labrad/qubits/QubitContext.java
pomalley: Oops, misread it, sramDirty gets set in the overload.

dac_calibration_server.py
```diff
@@ -259,90 +240,116 @@ def dac(self, c, dac):
...
+            print 'No deconv on board ' + c['Board']
+        returnValue(corrected)
+
+    @setting(33,
+        'Correct Analog FT',
+        data=['*v: Single channel data'],
+        average_ends='b',
+        dither='b',
+        returns=['*i: Single channel DAC values'])
+    def correct_z_ft(self, c, data, average_ends=False, dither=False):
```
pomalley: I much prefer it when the python function name matches the LabRAD setting name.

pomalley:
Let me see if I have this right:

* The IQ boards can have their ends zeroed, and the analog boards can have theirs averaged.
* The default for the qubit sequencer is to do this, and the default for the deconv server is to not do it.
* The analog boards can also be dithered; the default is to not do this.
* In current master, the current default is to do all of the above (dithering, averaging, zeroing).
* There is no qubit sequencer setting to turn dithering on/off.

Since we don't have tests for this I'm hesitant to LGTM until it's been tested on a fridge (has it?). Furthermore, I'm also hesitant to change the default behavior without any changes made in the qubit stack to accommodate this. However, if it really is critical then maybe we should go ahead.

pomalley:
Also, I think the docstrings in the deconv server need to be updated to indicate the dither/average/zero arguments.

DanielSank:
@pomalley your bullet points are all correct as of this writing.

If "tested on a fridge" means "measure qubits", then no, it has not been tested. Please volunteer to do this since you have working qubits :-)

If "tested on a fridge" means "used the deconvolution server to make waveforms and run experiments" then the answer is yes.

JulianSKelly:
LGTM.

DanielSank:
@pomalley The qubit stack only needs to change in the qubit sequencer, and I think a9774911a covers it.

ejeffrey:
Can you remove the set_border_values @setting?  It doesn't do anything
except print a warning message to the console where nobody will see it.  It
isn't called in fpgaseqTransmon any more, @pomalley can you check the qubit
sequencer?

ejeffrey:
LGTM

JulianSKelly:
Did this require sister changes to fpgaseqTransmon to not break master?

DanielSank:
> Did this require sister changes to fpgaseqTransmon to not break master?

I should not, since the qubit sequencer does the right thing by default. Since you LGTM'd it I thought that meant you ran it on Vince.

PR #110, opened 2015-04-21
Remove import_parameters setting on data_vault server.

maffoo:
This setting imports a bunch of parameters from the registry by using the context of the given request and assuming that the same context has been set up with the registry. This sort of inter-server context synchronization is a bad idea and needs to go away. This setting is not actually being used, instead we usually just set parameters on our own using the registry wrapper, which is a much more sane way to go. Hence, removing this should be low impact.

### DIFF
--- a/data_vault.py
+++ b/data_vault.py
@@ -1147,68 +1147,6 @@ class DataVault(LabradServer):
         if len(params):
             return params
 
-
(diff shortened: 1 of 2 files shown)


ejeffrey:
LGTM

PR #113, opened 2015-04-23
Deconv server and qubit sequencer: added correction for reflections in Z control lines

rb25:
Added correction for reflections in the Z (analog board) line. Can work in addition to settlingRates. Changes are in correction, dac_calibration, and the qubit server.


ghzdac/correction.py
```diff
@@ -1110,6 +1130,9 @@ def DACifyFT(self, signal, t0=0, n=8192, offset=0, nfft=None, loop=False,
         decayRates = np.array([x['GHz'] for x in self.decayRates])
         decayAmplitudes = self.decayAmplitudes
 
+        reflectionRates = np.array([x['GHz'] for x in self.reflectionRates])
```
DanielSank: This looks funky. reflectionRates should be a `ValueArray` in which case the iteration is not necessary.

ghzdac/correction.py
```diff
@@ -1165,10 +1188,17 @@ def DACifyFT(self, signal, t0=0, n=8192, offset=0, nfft=None, loop=False,
                 #decay times:
                 # add to qubit registry the following keys:
                 # settlingAmplitudes=[-0.05]  #relative amplitude
-                # settlingRates = [0.012]    #rate is in GHz, (1/ns)
+                # settlingRates = [0.01]    #rate is in GHz, (1/ns)
                 if np.alen(decayRates):
```
DanielSank: This conditional looks odd; should be something like

```python
if decayRates is not None:
...
```

rb25: Max wrote it this way. It's because if there is no settling it's an empty list, so "if decayRates is not None" wouldn't work.

ghzdac/correction.py
rb25: Yeah, the deconv server doesn't do units, so I guess this is how units were dealt with. I found it like this for the settling part, copied that for the reflection part.

DanielSank: Not sure what it means that the server doesn't do units since the code has `x['GHz']`. Wherever `reflectionRates` is set perhaps we can make that a `ValueArray` instead of a list of `Value`, since that's what pylabrad unflattens `*v[Hz]` to anyway.

ghzdac/correction.py
DanielSank: I see, makes sense.

ghzdac/correction.py
rb25: Yes, but all the variables inside the server (time, frequency etc) don't have labrad units. Internally, time is in ns, freq in GHz, I don't want to upset that right now

ghzdac/correction.py
```diff
@@ -1165,10 +1188,17 @@ def DACifyFT(self, signal, t0=0, n=8192, offset=0, nfft=None, loop=False,
...
-                    precalc /= (1.0 + np.sum(decayAmplitudes[:, None] * freqs[None, :] / (freqs[None, :] + decayRates[:, None]), axis=0))
+                    precalc /= (1.0 + np.sum(decayAmplitudes[:, None] * 2j*np.pi*freqs[None, :] / (2j*np.pi*freqs[None, :] + decayRates[:, None]), axis=0))
+
+                #reflections:
+                # add to qubit registry the following keys:
+                # reflectionAmplitudes=[0.05]  #relative amplitude
+                # reflectionRates = [0.01]    #rate is in GHz, (1/ns)
+                if np.alen(reflectionRates):
+                    for rate,amplitude in zip(reflectionRates,reflectionAmplitudes):
+                        precalc /= (1.0 - amplitude) / (1.0-amplitude*np.exp(-2j*np.pi*freqs/rate))
```
maffoo: For decay correction, the contributions are summed and then precalc is divided by the sum. Here, you're dividing precalc by the contribution from each successive reflection term. Why the difference?

ghzdac/correction.py
```diff
@@ -1165,10 +1188,17 @@ def DACifyFT(self, signal, t0=0, n=8192, offset=0, nfft=None, loop=False,
                 #decay times:
                 # add to qubit registry the following keys:
                 # settlingAmplitudes=[-0.05]  #relative amplitude
-                # settlingRates = [0.012]    #rate is in GHz, (1/ns)
+                # settlingRates = [0.01]    #rate is in GHz, (1/ns)
                 if np.alen(decayRates):
-                    freqs = 2j*np.pi*freqs
```
maffoo: seems like it would still be nice to precompute this; maybe something like:
```python
if np.alen(decayRates) or np.alen(reflectionRates):
    i_pi_freqs = 2j*np.pi*freqs
    if np.alen(decayRates):
        precalc /= ...

    if np.alen(reflectionRates):
        for rate, amplitude in zip(reflectionRates, reflectionAmplitudes):
            precalc /= (1.0 - amplitude) / (1.0 - amplitude * np.exp(-i_pi_freqs / rate))
```

ghzdac/correction.py
rb25: This is a good one. For deconvolution, you divide by the impulse response function.
Suppose X is an ideal pulse, H the impulse response of a piece of cable (with reflection, settling etc). To get X at the end of the cable you need to send Y = X/H. 

While dividing is the fully correct approach, you can approximate the settling case by adding in a particular way. This is what Max did. It should be fine as long as the amplitudes are around a few percent. I can fix it, but it won't matter in practice.

ghzdac/correction.py
rb25: It actually gets precomputed. The 'precalc' array holds the board response, filter, settling and reflection. It is computed once. Only when reflection or settling is changed, precalc is emptied, and then recalculated, see line 1177:
if np.alen(self.precalc) != nrfft:

maffoo: I was referring just to the subexpression `2j*np.pi*freqs`, which previously was being computed once and then used twice in the settling response function. The way you've written it, this subexpression would be reevaluated several times in both the settling and reflection terms. Not a big difference either way, though.

ghzdac/correction.py
maffoo: It makes sense to me that in general you want to convolve the response functions, but I guess it depends on how you measure the reflection response. If you measure two reflections peaks in the time domain, say, then you get a time-domain response like `X(t) + a*X(t-t_a) + b*X(t-t_b)` which fourier transforms to something like `x(f) * (1 + a*e(-2pi f t_a) + b*e(-2pi f t_b))`, so now to correct for this you want to divide by the sum of the two reflection terms instead of the product. Obviously, I'm playing fast and loose with signs here, but I guess what I'd like to see here is some more detail of how were measuring and modeling the reflection response, so that we can be sure that the correction is being applied in a manner consistent with those measurements.

DanielSank: +1

rb25: Think about echo, with reflections you have an endless sum of copies of the original signal with decreasing amplitude:
f(t) ->   (1-a) Sum_k=0^\infty (a^k f(t-kT) )
with T the roundtrip time, and a the reflection coefficient.

ghzdac/correction.py
rb25: Good point. Will do

ghzdac/correction.py
maffoo: That's a good explanation; can you add some comments about the corrections that are being applied in the code? Also, a comment about how to measure the coefficients to be applied as corrections.

rb25: Yes, I'll definitely do that. I'll also update the doc on the twiki to document it nicely. It's always best to put up a figure.

maffoo:
This commit includes an updated qubit server jar, but does not bump the version number. We should bump the version as well. I can help make that change if you'd like. Otherwise, LGTM

rb25:
Yeah we should. Peter's on it.

pomalley:
Here's the deal: I rebased this branch on master and pushed to u/pomalley/reflectionrates. Could you both take a look and make sure things make sense? There were some conflicting changes to do with the dithering/setting ends to 0 stuff.

maffoo:
@pomalley, it would probably be easiest if you just open a new pull request on your branch. You can reference this one and then close it.

pomalley:
Closing this in favor of the rebase here: #133.

PR #122, opened 2015-04-29
Qubit Server changes for the jump table.

pomalley:
Here are the changes to the Qubit Sequencer for the jump table.

There are a couple important changes:

* The memory sequence commands have been factored out of the FPGA classes.
* We now have the `Controller` class which has `MemoryController` and `JumpTableController` subclasses.
* The FastBiasChannel has been refactored into FastBiasSerialChannel and FastBiasFPGAChannel.
* The `FastBiasChannelBuilder` will now build a `FastBiasSerialChannel` if the registry indicates that the fastbias card is to be controlled through the DC Rack server.
* There are several new settings added to QubitContext.java to accomodate this.

I've also included the diff file that this was generated from. That will be removed for the final commit.

See also: #123.


DanielSank:
Hey Peter, there's a .diff file in there. Is this intentional?

pomalley:
I put it in there for reference. I don't know, just in case someone wants to try applying it again or something. I am planning to remove it from the actual final commit.

Qubit Server/dev/src/main/java/org/labrad/qubits/Constants.java
```diff
@@ -35,7 +35,7 @@
   public static final String REGISTRY_SERVER = "Registry";
   public static final String HITTITE_SERVER = "Hittite T2100 Server";
   public static final String ANRITSU_SERVER = "Anritsu Server";
-  public static final String DC_RACK_SERVER = "DC Rack";
+  public static final String DC_RACK_SERVER = "DC Rack Server";
```
maffoo: I don't see the actual server name change either here or in the `ghz_fpga_server` changes; is this just not used currently?

Qubit Server/dev/src/main/java/org/labrad/qubits/FpgaModelDac.java
```diff
@@ -1,27 +1,28 @@
...
 import org.labrad.qubits.proxies.DeconvolutionProxy;
 import org.labrad.qubits.resources.DacBoard;
 
-import com.google.common.base.Preconditions;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Future;
```
maffoo: Is this intellij just reordering imports? We should agree on an ordering and set up all the IDEs to do it the same way to avoid superfluous diffs; my preference is just all the imports in one block, sorted alphabetically.

Qubit Server/dev/src/main/java/org/labrad/qubits/FpgaModelDac.java
```diff
@@ -51,156 +58,64 @@ public DacBoard getDacBoard() {
...
-   * Check that the timer status of this board is ok, namely that the timer
-   * has been started at least once and stopped as many times as it has been
-   * started.  This ensures that all boards will be run properly.
-   */
-  public void checkTimerStatus() {
-    Preconditions.checkState(isTimerStarted(), "%s: timer not started", getName());
-    Preconditions.checkState(isTimerStopped(), "%s: timer not stopped", getName());
+  private int startDelay = 0;
+  public void setStartDelay(int startDelay) {
+	  this.startDelay = startDelay;
```
maffoo: indentation looks funny. tabs?

Qubit Server/dev/src/main/java/org/labrad/qubits/FpgaModelDac.java
```diff
@@ -366,7 +199,8 @@ public boolean hasDualBlockSram() {
    * @return
    */
   public long[] getSramDualBlock1() {
-    Preconditions.checkState(sramCalledDualBlock, "Sequence does not have a dual-block SRAM call");
+    MemoryController memoryController = (MemoryController)controller;
```
maffoo: should use the `getMemoryController` method here.

Qubit Server/dev/src/main/java/org/labrad/qubits/FpgaModelDac.java
```diff
@@ -383,7 +217,8 @@ public boolean hasDualBlockSram() {
    * @return
    */
   public long[] getSramDualBlock2() {
-    Preconditions.checkState(sramCalledDualBlock, "Sequence does not have a dual-block SRAM call");
+    MemoryController memoryController = (MemoryController)controller;
```
maffoo: `getMemoryController` here too

Qubit Server/dev/src/main/java/org/labrad/qubits/QubitContext.java
```diff
@@ -449,11 +439,11 @@ public void mem_delay(@Accepts("v[us]") double delay) {
   @Setting(id = 321,
     name = "Mem Delay Single",
     doc = "Add a delay to a single channel.")
-  public void mem_delay_single(@Accepts({"(s v[us])", "((ss) v[us])"}) Data command) {
-    FastBiasChannel ch = getChannel(command.get(0), FastBiasChannel.class);
-    double delay_us = command.get(1).getValue();
-    getExperiment().addSingleMemoryDelay(ch.getFpgaModel(), delay_us);
-    memDirty = true;
+    public void mem_delay_single(@Accepts({"(s v[us])", "((ss) v[us])"}) Data command) {
```
maffoo: this indentation is inconsistent with the other settings

Qubit Server/dev/src/main/java/org/labrad/qubits/QubitContext.java
```diff
@@ -507,6 +497,57 @@ public void mem_sync_delay() {
...
+
+
+  //
+  // Jump Table
+  //
+
+  @Setting(id = 370,
+           name = "Clear Jump Table",
+           doc = "Clear the jump table commands in this context.")
+  public void clear_jump_table() {
```
maffoo: group jump table commands by name prefix: "Jump Table Clear", "Jump Table Add Entry", "Jump Table Set Counters"

Qubit Server/dev/src/main/java/org/labrad/qubits/channels/FastBiasSerialChannel.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  public boolean hasSetupPacket() {
+    return configured;
+  }
+
+  public SetupPacket getSetupPacket() {
+    Preconditions.checkState(hasSetupPacket(), "Cannot get setup packet for " +
+                    "channel '%s': it has not been configured.", getName());
+    int dac_num, rc_time_constant;
```
maffoo: prefer camelCase names in java

Qubit Server/dev/src/main/java/org/labrad/qubits/jumptable/JumpTable.java
```diff
@@ -0,0 +1,85 @@
...
+    this.counters = counters;
+  }
+
+  public boolean equals(JumpTable other) {
+    if (entryNames.size() != other.entryNames.size()) {
+      return false;
+    }
+    for (int i=0; i<entryNames.size(); i++) {
+      if (!(entryNames.get(i).equals(other.entryNames.get(i)) &&
+            entryArguments.get(i).equals(other.entryArguments.get(i)))) {
```
maffoo: I don't think the jlabrad `Data` class implements equals properly. We fixed this for trigger tables by comparing the pretty-printed string representation instead: https://github.com/martinisgroup/servers/commit/b64aa3b4f537a64267bd3595a3c5e37ccfaf234a

Note that this is fixed in scalabrad, where the Data class does implement equality.

maffoo:
This piece LGTM as it stands, although philosophically I would prefer to shift some of the responsibility from what is now in pyle to the qubit sequencer. For example, to run a dual-block sequence in the past you would upload two SRAM blocks, then tell the sequencer to run them with a certain delay. The natural (in my head) extension of that for the jump table would be to upload all your blocks and then tell the sequencer how to run them with cycles, idles, etc. Much of the work that is being spread across the jump table compiler and fpgaseqTransmon would then happen here instead. The main complication I could see with this approach would be dealing with delays, since at least the sub-nanosecond (or sub-4-nanosecond) part of the delay would need to be dealt with in pyle when evaluating envelopes). But I think that complexity is manageable, and would simplify the pyle code a good deal.

I know people are less comfortable modifying the qubit sequencer because of less familiarity with java, the extra dev setup and the extra deployment steps, but I think these things will get better. When we convert to scala, there will just be a lot *less code* here, which I think will make understanding the code and making changes easier. We're working on making the deployment of new versions easier so that will be less of an impediment as well. However, given that those process improvements are not here *yet*, perhaps it's best to push ahead with the approach you've taken at the moment, and we'll shift the responsibilities over time. Thoughts?

Qubit Server/dev/src/main/java/org/labrad/qubits/Constants.java
pomalley: I believe the name of the server has been "DC Rack Server" for a while, and was never changed (and evidently never used) in the qubit sequencer.

Qubit Server/dev/src/main/java/org/labrad/qubits/FpgaModelDac.java
pomalley: Yeah, this is intellij. I haven't looked into how it does it but your preference sounds fine to me.

Qubit Server/dev/src/main/java/org/labrad/qubits/jumptable/JumpTable.java
pomalley: Yikes!

DanielSank:
In terms of being uncomfortable modifying the qubit sequencer, I personally would like to remedy that. The problem is that learning new languages (and build systems) is hard without really concerted effort.

Matthew, can I convince you to give us a 1 hour scala lesson at some point? If nothing else this will force us to come up with a somewhat standard dev setup that everyone understands.

maffoo:
@DanielSank, yes, gladly. I have on my todo list to write down the instructions for setting up sbt and building/running scala, which should help people get started. I'd be happy to go through it with people once that's ready.

pomalley:
I haven't made any changes on this branch yet, but I'll remind myself here that I need to change the calls to the FPGA server to reflect the renamed JT calls.

Qubit Server/dev/src/main/java/org/labrad/qubits/jumptable/JumpTable.java
pomalley: Turns out this isn't used. Back when, the qubit sequencer compared the JT commands you gave for different boards to make sure they were all equal. That is no longer the case as the JT commands are global to the experiment. So I will delete this function.

pomalley:
I pushed new commits to fix the issues @maffoo brought up, as well as to account for the name changes in the fpga server. Let me know what y'all think!

pomalley:
I'd like to rebase this on current master and force push--any objections?

maffoo:
no objection here.

pomalley:
OK, done.

pomalley:
I'm planning on merging in the FPGA server side of this on Saturday and I'd like to do this PR as well. @maffoo you already said "LTGM as it stands" so I'm going to go with that. :-)

PR #124, opened 2015-04-29
Convert run_sequence result to single numpy array before flattening.

maffoo:
Previously, the result from run_sequence was being returned as a python list of numpy arrays, so that the special flattening code for numpy arrays was being bypassed. With this change, @JulianSKelly and I saw roughly a factor 2x increase in data throughput under load.

### DIFF
--- a/GHzDACs/ghz_fpga_server.py
+++ b/GHzDACs/ghz_fpga_server.py
@@ -177,7 +177,7 @@ cmdTime_cycles does not properly estimate sram length
 ### BEGIN NODE INFO
 [info]
 name = GHz FPGAs
-version = 4.0.2
+version = 4.0.3
 description = Talks to DAC and ADC boards
 
 [startup]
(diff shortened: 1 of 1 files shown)


jwenner:
Would it be faster to have this numpy-izing just before run returns data at https://github.com/martinisgroup/servers/blob/u/maffoo/ghz-fpga-speedup/GHzDACs/ghz_fpga_server.py#L783 so that you aren't passing the list of arrays around even within the same file?

maffoo:
@jwenner, passing lists of arrays in python is not expensive, it's just that when we try to flatten this, pylabrad doesn't use the optimized numpy array flattening code.

DanielSank:
For future readers who are confused by this:

The problem is that if we pass a python list of numpy arrays to the flattening code, pylabrad iterates at all levels. It iterates through the list, *and* for each numpy array in the list it iterates through each element of the numpy array. This is super-duper inefficient. It would be better if we detected when a list contains numpy arrays and used the efficient flattening of those arrays, but that's just not supported. If we instead make sure to pass in a numpy array (with one extra index) instead of a list of arrays, then pylabrad uses efficient flattening of that array without iterating over every single value.

PR #125, opened 2015-05-01
Change ghz_fpga_server to always read a list of packets.

maffoo:
The delphi direct ethernet server changes the 'shape' of packet responses when you try to read one versus multiple packets. In particular, if you call `.read()` it returns a single packet as `(ssis)`, but if you call `.read(1)` it returns a list of one packet as `*(ssis)`, which obviously is the same type used for larger numbers of packets as well. By changing all the read calls to specify the number of packets explicitly, we always get the same type back, which normalizes the response handling. In the new scala version of the direct ethernet server, this allows us to simplify the implementation since we can just specify a single return type.

### DIFF
--- a/GHzDACs/Cleanup/fpga.py
+++ b/GHzDACs/Cleanup/fpga.py
@@ -176,10 +176,10 @@ class FPGA(DeviceWrapper):
         p.write(regs.tostring())
         if readback:
             p.timeout(timeout)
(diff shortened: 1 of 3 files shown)


DanielSank:
LGTM

ejeffrey:
LGTM

PR #126, opened 2015-05-04
Remove .doc version of ADC documentation. (turbo easy)

DanielSank:
The original file still lives in John's directory so we
don't need it duplicated here now that we have the .md
version.


ejeffrey:
LGTM

PR #132, opened 2015-05-06
Add scala version of direct ethernet server.

maffoo:
Scala version of the direct ethernet server. This has been tested on Vince and is currently running taking data, with lower cpu usage than the delphi server. Note that I've set up the Node file to list this as "Scala Direct Ethernet" but it actually runs with the same instance name as the delphi server, so you can have both of these be auto-detected by the node and just run the one you want. However, I haven't checked in the built jars here, so if you were to check this out without building, the node would detect the ini file but then not be able to run the server. We need to figure out what to do about binary servers in such cases; thoughts? In the meantime, wanted to get this in front of people so they can take a look at some scala. There are still a few things about the scalabrad API that I'd like to fix (for example, the way the `Server` subclass *and* the `ServerConnection` object have the context class as a type parameter is awkward), but things are already working nicely.


DanielSank:
Why don't we start using some of git's hook capabilities to help with these deployment issues?

maffoo:
@DanielSank, I'm not sure I see what you mean

DanielSank:
I know git has hooks for scripts to run when you commit. I'm not sure about fetch/pull but I bet there's something out there.

DanielSank:
* The server cannot be started from the node with the new Scala server.

* When I run the new server from cmd.exe it prints the following:
```
C:\labrad\servers\DirectEthernet-1.0.0-M1\bin>directethernet.bat
15:55:08.368 [main] WARN  org.pcap4j.util.PropertiesLoader - [org/pcap4j/pcap4j.
properties] Could not get value by org.pcap4j.af.inet, use default value: null
15:55:08.378 [main] WARN  org.pcap4j.util.PropertiesLoader - [org/pcap4j/pcap4j.
properties] Could not get value by org.pcap4j.af.inet6, use default value: null
15:55:08.378 [main] WARN  org.pcap4j.util.PropertiesLoader - [org/pcap4j/pcap4j.
properties] Could not get value by org.pcap4j.af.packet, use default value: null

15:55:08.378 [main] WARN  org.pcap4j.util.PropertiesLoader - [org/pcap4j/pcap4j.
properties] Could not get value by org.pcap4j.af.link, use default value: null
15:55:08.388 [main] INFO  org.pcap4j.core.Pcaps - 4 NIF(s) found.
15:55:08.398 [main] WARN  org.pcap4j.util.PropertiesLoader - [org/pcap4j/pcap4j.
properties] Could not get value by org.pcap4j.dlt.raw, use default value: null
15:55:08.418 [main] INFO  org.labrad.ServerConnection$ - LABRADNODE -> DR
15:55:09.168 [main] INFO  org.labrad.ServerConnection - Now serving...
```

maffoo:
Those log warnings you can ignore; we can dig into what's causing them, but they don't actually have any effect on the operation of the server. Latest commit fixes the node startup issue, and the autodetect problem I think is a bug in the ghz_fpga_server that happens when you restart the direct ethernet server while the ghz_fpga server is running: #134

maffoo:
Added a .nodeignore file so that the node will not scan this directory (see martinisgroup/pylabrad#100), which means that this source can live in the checkout with other servers that _should_ be scanned by the node and we can deploy the compiled binaries separately.

DanielSank:
Seems to work. I can't vouch for the code though so I'm not sure if this is an LGTM...

DanielSank:
@maffoo how on earth are we supposed to review this? Nobody else knows the language yet.

maffoo:
@DanielSank, sounds like a good plan. Please do take a look over the weekend to just read through some scala code, and we can go through it in detail next week.

DanielSank:
Deal :-)

DanielSank:
LGTM (whatever that's worth ;)

PR #136, opened 2015-05-09
Gracefully handle unavailable executionCount

### LINKED ISSUE #135: fpga server timeout recovery raises KeyError on 'executionCounter'
This is happening because the ADC class doesn't set `executionCounter` in the readback data.
I'm gunna just make it gracefully handle the case where `executionCounter` is not available.

Future work should implement the execution counter on any boards which support it!
I think this isn't implemented properly now out of laziness (probably laziness of past Dan).

DanielSank:
The server's recover from timeout strategy involves reading an execution counter form the boards. Unfortunately, not all boards support this and the server was failing on a `KeyError`.
Now we gracefully handle this case.

Tested running on Jules, got timeout, server reported execution counts and carried on!

Fixes #135.

### DIFF
--- a/GHzDACs/ghz_fpga_server.py
+++ b/GHzDACs/ghz_fpga_server.py
@@ -839,11 +839,15 @@ class BoardGroup(object):
                 # processReadback as a module level function in dac.py,
                 # whereas in adc.py it's a staticmethod of the ADC class.
                 if isinstance(runner, dac.DacRunner):
-                    count = runner.dev.\
-                            processReadback(ans[0][3])['executionCounter']
+                    count = runner.dev.processReadback(ans[0][3]).get(
+                        'executionCounter',
+                        None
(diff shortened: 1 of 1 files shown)


GHzDACs/ghz_fpga_server.py
```diff
@@ -859,9 +863,15 @@ def timeoutReport(self, runners, results):
         """Create a nice error message explaining which boards timed out."""
         lines = ['Some boards failed:']
         for runner, (success, result) in zip(runners, results):
+            if runner.executionCount is None:
+                executionCount = 'not available'
+            else:
+                executionCount = runner.executionCount
```
maffoo: I see this was already merged, but for future reference this could be written as
```python
executionCount = getattr(runner, 'executionCount', 'not available')
```

Issue #99, opened 2015-04-16
FPGA Error Message Unclear for Board not in BoardGroups


jwenner:
We tried to run fpgaTest.average (which calls fpga.run_sequence) with only the DAC (i.e., not the ADC) in boardGroups. When we did this, the error should be something about a board not in boardGroups. However, we instead got a timeout error and a labrad.types.Error.

The particular error message in the fpgaTest IPython session was:
```
In [15]: fpgaTest.average(s,cxn,3,True)
---------------------------------------------------------------------------
Error                                     Traceback (most recent call last)
<ipython-input-15-53c37bbfb41c> in <module>()
----> 1 fpgaTest.average(s,cxn,3,True)

C:\pyle\pyle\scripts\fpgaTest\fpgaTest.pyc in average(sample, cxn, reps, plot, n
ame, dacsAdcs)
   1086     fpga.daisy_chain(daisyChainList)
   1087     fpga.timing_order([adc['_id']])
-> 1088     data = fpga.run_sequence(reps, True)[0]
   1089     if plot:
   1090         plt.figure()

C:\pylabrad\labrad\client.pyc in __call__(self, *args, **kw)
     62         if wrap:
     63             future.addCallback(lambda resp: resp[0][1])
---> 64         return future.wait() if wait else future
     65
     66     # data to be loaded on demand

C:\pylabrad\labrad\backend.pyc in wait(self)
    397                 # If any Future in the queue fails,
    398                 # we immediately bail.
--> 399                 result.raiseException()
    400             else:
    401                 for func, args, kw in f.callbacks:

C:\Python27\lib\site-packages\twisted\python\failure.pyc in raiseException(self)


Error: (0) [GHz FPGAs] Remote Traceback (most recent call last):
  File "C:\pylabrad\labrad\server.py", line 262, in handleRequest
    result = yield setting.handleRequest(self, c.data, data)
  File "C:\Python27\lib\site-packages\twisted\internet\defer.py", line 1105, in
_inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "C:\Python27\lib\site-packages\twisted\python\failure.py", line 389, in t
hrowExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "C:\pylabrad\servers\GHzDACs\ghz_fpga_server.py", line 1449, in run_seque
nce
    timingOrder)
  File "C:\Python27\lib\site-packages\twisted\internet\defer.py", line 1105, in
[...]

PR #138, opened 2015-05-09
Remove exception raised by recoverFromTimeout.

### LINKED ISSUE #137: fpga server recover from timeout raises un-necessary exception
`recoverFromTimeout` raises an exception if pinging the boards fails. This exception is then unhandeled and leads to sadness<sup>TM</sup>.
This exception is un-necessary since we already have a finally clause to clear the packet buffers and send appropriate triggers.
Therefore, we should just remove it.

DanielSank:
(Making PR so folks can tell me if I'm being high with this strategy)

We explicitly silence exceptions occuring during the board pings. This is ok because we have a finally clause to clear the packet buffers in the direct ethernet server.

Also increase number of retries for timeout errors.

The only thing left to do here (I think) is improve the error logging e.g. by adding time stamps.

Fixes #137.

### DIFF
--- a/GHzDACs/ghz_fpga_server.py
+++ b/GHzDACs/ghz_fpga_server.py
@@ -254,7 +254,8 @@ I2C_END = 0x400
 
 class TimeoutError(Exception):
     """Error raised when boards timeout."""
-    
+
+
 class BoardGroup(object):
     """Manages a group of GHz DAC boards that can be run simultaneously.
     
@@ -1474,18 +1475,19 @@ class FPGAServer(DeviceServer):
                 userpath = os.path.expanduser('~')
                 logpath = os.path.join(userpath, 'dac_timeout_log.txt')
                 with open(logpath, 'a') as logfile:
-                    print 'attempt %d - error: %s' % (attempt, err)
-                    print >>logfile, 'attempt %d - error: %s' % (attempt, err)
+                    t = timeString()
+                    msg = '{}: attempt {} - error: {}'.format(t, attempt, err)
+                    print(msg)
+                    logfile.write(msg+'\n')
                     if attempt == retries:
-                        print 'FAIL!'
-                        print >>logfile, 'FAIL!'
+                        logfile.write('FAIL\n')
                         #TODO: notify users via SMS
                         raise
                     else:
-                        print 'retrying...'
-                        print >>logfile, 'retrying...'
+                        print('retrying...')
+                        logfile.write('retrying...')
                         attempt += 1
-    
+
     @setting(52, 'Daisy Chain', boards='*s', returns='*s')
     def sequence_boards(self, c, boards=None):
         """
--- a/GHzDACs/ghz_fpga_server.py
+++ b/GHzDACs/ghz_fpga_server.py
@@ -850,8 +850,8 @@ class BoardGroup(object):
                     )
                 runner.executionCount = count
             except Exception as e:
+                print("Attempt to read execution counts failed: ")
                 print e
-                raise Exception('Recover from error failed')
             # Finally, clear the packet buffer and send trigger if this was a
             # failed board
             finally:


GHzDACs/ghz_fpga_server.py
```diff
@@ -880,7 +880,7 @@ class FPGAServer(DeviceServer):
     """Server for GHz DAC and ADC boards.
     """
     name = 'GHz FPGAs'
-    retries = 5
+    retries = 100
```
maffoo: wow, 100 retries? is our error rate really that bad?

GHzDACs/ghz_fpga_server.py
```diff
@@ -850,8 +850,8 @@ def recoverFromTimeout(self, runners, results):
                     )
                 runner.executionCount = count
             except Exception as e:
+                print("Attempt to read execution counts failed: ")
```
maffoo: We should be consistent in using or not using parens with print. I like doing `from __future__ import print_function` and using parens everywhere, but we should probably make that change in one go for the whole file. Until then, omit.

maffoo:
I agree that failing when the execution count reading fails is not right. Should just move on and try to recover. However, we need to know the rate at which this is happening as we don't want to mask real underlying problems. It seems to me that if it takes 100 retries to recover, the system is in a bad state, and making so many attempts will just delay finding out about it.

DanielSank:
I picked 100 because I figured that if the system really falls apart then it will burn through the 100 tries and still fail, thus notifying the user that something is really wrong.

I guess what you're suggesting is that a long string of e.g. 50 retries is also worthy of user notification. I can make a little buffer which halts the program if some statistical metric goes above a certain threshold. Is that what you had in mind?

GHzDACs/ghz_fpga_server.py
DanielSank: I try to use parens everywhere. Is there a downside to that?

GHzDACs/ghz_fpga_server.py
DanielSank: I don't know. I want to make this number large and improve the logging so we can find out what's going on.

DanielSank:
Ok now I think this is actually ready to be reviewed.

ejeffrey:
So earlier we were having a problem on Jules that looked like it was never succeeding in pinging the board on a timeout, and then causing the recovery to fail completely, and taking the whole scan down with it.  This sounds like the recovery sequence is just broken.  Going from 5 to 100 retries doesn't really sound like a fix, our packet loss rates are tiny.  If this is failing, something else is going on.  Either the boards don't listen for ping packets while waiting to run, or the network link is just dropping for a couple of seconds, or something like that.

Can you deliberately generate a timeout and see if the code is actually capable of recovering from a timeout successfully?

DanielSank:
> So earlier we were having a problem on Jules that looked like it was never succeeding in pinging the board on a timeout

I did not know that.

> Can you deliberately generate a timeout and see if the code is actually capable of recovering from a timeout successfully?

Not right now because a long scan is running. I'll do that when I can though.

DanielSank:
@maffoo @ejeffrey What needs to happen to make this acceptable? Do you want me to just go back to a smaller number of retries?

GHzDACs/ghz_fpga_server.py
```diff
@@ -254,7 +254,17 @@ def waitForkey():
...
-    
+
+try:
+    import datetime
+    def time_str():
+        return datetime.datetime.now().strftime("%Y-%m-%d %H:%M")
+
+except ImportError
+    def time_str():
+        return ''
```
maffoo: how could this import possibly fail? `datetime` is in the standard library, no?

maffoo:
Other than the change to `retries`, I'm fine with the rest (though the import guard around datetime seems unnecessary). The problem is that timeouts are slow, so waiting for things to timeout 100 times will take a long time before the failure propagates back up. And actual timeouts due to packet loss are statistically rare, so it seems like 5 retries is plenty (even if packet loss happened 1% of the time, the probability of 5 failed attempts in a row is 1e-10). Maybe we could go to 8 or 10, but anything beyond that will just delay notifying the user when there is some other more serious problem in the system.

pomalley:
I agree with @maffoo on the retries. If there is anything other than statistical raw ethernet packet loss causing the timeouts, I would guess that there is something wrong that needs user intervention.

jwenner:
In the print statement near the exception message, can we add something about checking the daisy chain? In my experience, daisy chain errors are one of the main times that we have timeouts. This would then fix #99 as well.

DanielSank:
@maffoo I got rid of the import guard and went back to 5 retries. Ok now?

maffoo:
yeah, LGTM

PR #141, opened 2015-05-11
Do not assume case-insensitivity when building registry paths.

maffoo:
When building registry paths in some instance were were lower-casing the directory names and assuming the registry was case-insensitive, which was an artifact of the registry being a thin wrapper around the windows filesystem. In the new delphi implementation, registry paths are case sensitive. We had to make these changes to get the scala manager and registry proxy working on vince, er Vince.

### DIFF
--- a/lakeshore370.py
+++ b/lakeshore370.py
@@ -151,14 +151,10 @@ class RuOxWrapper(GPIBDeviceWrapper):
         yield self.reloadSettleTime(path)
     
     def getRegistryPath(self):
-        """Get a registry path suitable for registry.cd
-        
-        Warning - the registry is controlled by the same program as the manager
-        which runs in Windoze. This means that paths are case insensitive.
-        """
+        """Get a registry path suitable for registry.cd"""
         path = ['', 'Servers', 'Lakeshore 370']
-        gpibServerName, addr = self.name.split(' - ')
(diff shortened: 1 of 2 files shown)


lakeshore370.py
```diff
@@ -151,14 +151,10 @@ def loadDeviceInformation(self):
...
-        """Get a registry path suitable for registry.cd
-        
-        Warning - the registry is controlled by the same program as the manager
-        which runs in Windoze. This means that paths are case insensitive.
-        """
+        """Get a registry path suitable for registry.cd"""
         path = ['', 'Servers', 'Lakeshore 370']
-        gpibServerName, addr = self.name.split(' - ')
-        nodeName = gpibServerName.split(' ')[0].lower()
+        gpibServerName, addr = self.name.split(' - ', 1)
```
DanielSank: What's with the `, 1`?

maffoo: It means split at most one time. Otherwise, if `addr` itself contains the string ' - ' then it would be split again and we'd get three (or more) segments, instead of the desired two.

DanielSank: Testing this in the interactive session, I don't get it.

```python
s = '123-456-789'
s.split('-', -1)
['123', '456', '789]
```

DanielSank:
LGTM

Issue #149, opened 2015-05-13
Backwards compatibility broken on old dataset with Units None


joshmutus:
When opening an old dataset with the new units branch I get the following traceback:
  File "data_vault.py", line 504, in getPar
    data = T.evalLRData(S.get(sec, 'Data', raw=True))
  File "/Library/Python/2.7/site-packages/labrad/types.py", line 394, in evalLRData
    return eval(s)
  File "<string>", line 1, in <module>
    
  File "/Library/Python/2.7/site-packages/labrad/units.py", line 190, in __new__
    raise RuntimeError("Cannot construct WithUnit with unit=None.  Use correct units, or use a float")
exceptions.RuntimeError: Cannot construct WithUnit with unit=None.  Use correct units, or use a float
 [payload=None]

This dataset contains the parameters
[Parameter 14]
label = ParampBiasVoltage
data = Value(0.90000000000000302, None)

maffoo:
Oh man, that's bad. Good find. I have been worried about this text-based storage of data values for some time, precisely because it's so brittle and hard to maintain compatibility. In my update to the data vault, I changed the parameter saving to instead save a base64-encoded version of the raw labrad data to avoid this kind of thing (#24). Note that I also added a binary data format in that branch, but that's likely to be superseded by @ejeffrey's work on hdf5 support. However, we may want to strip out the binary data format and merge the rest, to avoid issues like this cropping up in the future. Of course, that won't fix the issue here, which might require that we manually migrate the old parameters to a compatible format, like `Value(0.9, '')` or even just `0.9`.

ejeffrey:
Yeah, I am already working on a fix for this in main that will just special
case 'None' values and convert them to dimensionless units.

Evan

joshmutus:
So I ran the data_vault from Evan's commit and I get a different error now:

Error: (0) [Data Vault] Remote Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/labrad/server.py", line 262, in handleRequest
    result = yield setting.handleRequest(self, c.data, data)
  File "data_vault.py", line 1050, in open
    dataset = session.openDataset(name)
  File "data_vault.py", line 397, in openDataset
    dataset = Dataset(self, name)
  File "data_vault.py", line 470, in __init__
    self.load()
  File "data_vault.py", line 514, in load
    self.parameters = [getPar(i) for i in range(count)]
  File "data_vault.py", line 505, in getPar
    data = T.evalLRData(datastr)
[...]

PR #150, opened 2015-05-13
Refactor data_vault and data_vault_multihead to share common code

maffoo:
This includes the refactoring changes from #24 minus the binary data backend, and with the addition of the parameter evaluation fix from #149. This will give us a much better foundation to build on going forward (e.g. for hdf5 support) since we will not have to maintain parallel versions of the single- and multi-headed versions of the data vault.

### DIFF
--- a/data_vault.py
+++ b/data_vault.py
@@ -41,44 +41,16 @@ timeout = 5
 
 from __future__ import with_statement
 
-from labrad import types as T, util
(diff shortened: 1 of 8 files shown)


maffoo:
@ejeffrey, you reviewed most of this already in #24, and your reservations there were about the binary backend, which is gone now, so this should be a pretty quick one.

ejeffrey:
LGTM

PR #160, opened 2015-05-19
Remove binary artifacts for qubit server

maffoo:
This one looks a lot worse than it is. The big change is to move the qubit server code from `Qubit Server/dev/` to `QubitServer/`. The extra subfolder is unnecessary because I've removed the binary artifacts which were in the `Qubit Server/lib/` directory. Instead, we'll distribute the built binary artifacts separately, as for the scala direct ethernet server. Note that before we merge this to master and deploy on any target hosts, we'll have to set up the binary distribution on those hosts.


DanielSank:
LGTM up to getting rid of that TODO.txt file.

maffoo:
I deployed the built artifacts on jules and vince so the next time the server is restarted, it'll pick up the binary code. Tested this on jules with no issues.

Issue #164, opened 2015-05-21
Convert Qubit Server to scala


maffoo:
We'd like to convert the qubit server to scala, to consolidate on a single JVM language and to take advantage of the performance improvements in the scalabrad api. Because this is a big change, we'll be taking it in stages, and will use this issue to track the overall progress. The plan is roughly as follows:

1. Translate chunks of files from java to scala. Doing this in chunks will keep the size of reviews small, and also allow us to maintain a working system at each stage. Because we'll be interoperating with java and scala, we won't be able to use a lot of the scala goodies. This will be a good thing, because we'll limit the changes to a very basic transliteration between the language, which will help people get used to reading scala. This will mean that the scala code may look awkward, but trust me, it gets better :)
2. Once everything is in scala, we can add in some of the scala goodies to simplify the code, use scala collections, etc. Again, we'll do this in small chunks so that reviews are manageable. Hopefully here some of the advantages of scala will start to become clear.
3. Convert from jlabrad to scalabrad. The scalabrad api will feel much more scala-y than the jlabrad api, so we'll be able to get rid of the remaining javaisms at this point and really clean up the code.

jwenner:
@maffoo, with #250 in, does this mean that this issue can be closed?

maffoo:
Yup, thanks @jwenner!

PR #165, opened 2015-05-21
Convert channeldata classes from java to scala

maffoo:
see the scala conversion plan: #164

Note that the PR includes separate commits to rename the files and then convert the code. If you look at the full change set, git doesn't recognize the files as renamed because so much changed. Instead, please just look at the last commit which changes the code after the rename. I will probably leave these as separate commits when merging to master so that we can track changes through the rename and we don't introduce a break in the file history.


DanielSank:
I gave this a first go-through, but I'd like to be more thorough. What was that resource you mentioned for learning Scala syntax?

maffoo:
Thanks for looking at this, @DanielSank. A few links that might be helpful: here's a really quick [tutorial](http://docs.scala-lang.org/tutorials/scala-for-java-programmers.html) on the scala site aimed at java programmers; twitter's [scala school](http://twitter.github.io/scala_school/) is a longer tutorial introduction to the language (they emphasize twitter libraries, but the language stuff is generally applicable); this [cheat sheet](http://techblog.realestate.com.au/java-to-scala-cheatsheet/) has lots of snippets in java and scala and might be a useful reference.

DanielSank:
Ah, that cheat sheet is nice.

maffoo:
@pomalley, are you interested in reviewing this? If so, when do you think you'll have time? I'd like to push ahead with the conversion so we can move on to feature work.

pomalley:
Yes I do want to review it, sorry I haven't gotten to it yet. I got distracted reading about scala. I'll try to get it done today. (And you should finish the batch qubit PR... ;-)

QubitServer/src/main/java/org/labrad/qubits/channeldata/AnalogData.scala
```diff
@@ -0,0 +1,9 @@
+package org.labrad.qubits.channeldata
+
+import org.labrad.qubits.channels.AnalogChannel
+
+trait AnalogData extends Deconvolvable {
+  def setChannel(channel: AnalogChannel): Unit
+  def getDeconvolved(): Array[Int]
```
pomalley: is it more standard to leave off the empty-parens? I guess the more standard thing tod do would be to name it `deconvolvedData` or something.

maffoo: yeah, you're correct on both counts. I have changes along those lines queued up for later, but I've left the names as they are for now because all the code that calls into these classes is still java.

QubitServer/src/main/java/org/labrad/qubits/channeldata/AnalogDataFourier.scala
```diff
@@ -0,0 +1,48 @@
+package org.labrad.qubits.channeldata
+
+import java.util.concurrent.Future
+
+import org.labrad.qubits.channels.AnalogChannel
```
pomalley: unused import

QubitServer/src/main/java/org/labrad/qubits/channeldata/AnalogDataFourier.scala
```diff
@@ -0,0 +1,48 @@
...
+        averageEnds,
+        dither
+    )
+    Futures.chain(req, new Function[Array[Int], Void] {
+      override def apply(result: Array[Int]): Void = {
+        deconvolvedData = result
+        setDeconvolved(true)
+        null
+      }
+    })
```
pomalley: I feel like there should be some sexy scala way to do this but I don't know enough yet to suggest anything. Also for now I'm glad you didn't do this in some sexy scala way because then I wouldn't know what's going on. :-)

QubitServer/src/main/java/org/labrad/qubits/channeldata/AnalogDataTime.scala
```diff
@@ -0,0 +1,68 @@
...
+import com.google.common.base.Function
+
+class AnalogDataTime(data: Array[Double], isDeconvolved: Boolean, averageEnds: Boolean, dither: Boolean) extends AnalogDataBase {
+
+  private var deconvolvedData: Array[Int] = null
+
+  if (isDeconvolved) {
+    deconvolvedData = data.map { x =>
+      (x * 0x1fff).toInt & 0x3fff
+    }
```
pomalley: well this is certainly nicer than the java way.

DanielSank: Yeah, the way you can just write out anonymous functions is sick.

QubitServer/src/main/java/org/labrad/qubits/channeldata/AnalogDataTime.scala
```diff
@@ -0,0 +1,68 @@
+package org.labrad.qubits.channeldata
+
+import java.util.concurrent.Future
+
+import org.labrad.qubits.channels.AnalogChannel
```
pomalley: again, unused import

QubitServer/src/main/java/org/labrad/qubits/channeldata/AnalogDataTime.scala
```diff
@@ -0,0 +1,68 @@
...
+      override def apply(result: Array[Int]): Void = {
+        deconvolvedData = result
+        setDeconvolved(true)
+        null
+      }
+    })
+  }
+
+  override def getDeconvolved(): Array[Int] = {
+    require(isDeconvolved(), "Data has not yet been deconvolved")
```
pomalley: I like that this is a Scala built in now, rather than google.something.preconditions.whatever.

QubitServer/src/main/java/org/labrad/qubits/channeldata/AnalogDataTime.scala
```diff
@@ -0,0 +1,68 @@
...
+    deconvolvedData
+  }
+}
+
+class AnalogDataTimeDacified(data: Array[Int]) extends AnalogDataBase {
```
pomalley: What? You don't need to put it in a new file?! ;-)

QubitServer/src/main/java/org/labrad/qubits/channeldata/AnalogDataFourier.scala
maffoo: yes, again, I have a later commit that converts all of the java futures to scala futures, which support this kind of chaining in a much cleaner way. Would it help if I just push my entire scala conversion branch so you could look ahead at all the commits? I don't want things to be overwhelming, but I'm happy to do it if you like.

pomalley: no no, i'm slowly making my way through this, this is a good first step

QubitServer/src/main/java/org/labrad/qubits/channeldata/IqDataBase.scala
```diff
@@ -0,0 +1,33 @@
+package org.labrad.qubits.channeldata
+
+import org.labrad.qubits.channels.IqChannel
+
+abstract class IqDataBase extends IqData {
+
+  private var channel: IqChannel = _
```
pomalley: what exactly does underscore mean here?

QubitServer/src/main/java/org/labrad/qubits/channeldata/AnalogDataTime.scala
maffoo: yup :)

scala completely decouples the file name from the class names, so you have to be a little bit careful to name things in a sane way, but it's very convenient to collect small related classes into a single file. I find it makes comprehending the larger structure of the code much easier than when every class is in a different file.

QubitServer/src/main/java/org/labrad/qubits/channeldata/IqDataBase.scala
maffoo: in a variable initialization expression like this, underscore means the default 'zero' value for the type in question. Here, that would be `null`. I should probably just be explicit here and write null, there's really nothing to gain from using underscore.

pomalley: OK, i was just confused because elsewhere (e.g. IqDataTime) it is explicitly set to null. I think we should go with whatever is idiomatic Scala. :)

pomalley:
Pending testing this out on a fridge, LGTM. I rather wish we had qubit sequencer tests...

maffoo:
@DanielSank and @pomalley, thanks for the reviews; will one of you have a few minutes to actually test this on a fridge? You'd just need to checkout the branch, cd into the QubitServer directory, and start it manually: `sbt "runMain org.labrad.qubits.QubitServer"`

DanielSank:
I think we can do this tomorrow. Will you be available for consultations?

maffoo:
yup, I'm around all day

DanielSank:
Seems to work. LGTM.

PR #167, opened 2015-05-22
Expand "~" and "~username" in data_vault.

### LINKED ISSUE #144: datavault should expand user directory when asking for vault path from command line
see maffoo/scalabrad#2

strj:
Allows the repository location to be specified with ~ and ~username
instead of just the full path. Fixes #144 and maffoo/scalabrad#2.


maffoo:
LGTM

PR #170, opened 2015-05-27
Fix accepted data type for analog IQ correction.

maffoo:
Fourier data should clearly be complex, not real as was being advertised.

### DIFF
--- a/dac_calibration_server.py
+++ b/dac_calibration_server.py
@@ -17,7 +17,7 @@
 ### BEGIN NODE INFO
 [info]
 name = DAC Calibration
-version = 1.2.0
+version = 1.2.1
 description = Calibrate sequences for the GHz DAC boards.
 
 [startup]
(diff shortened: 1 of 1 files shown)


dac_calibration_server.py
```diff
@@ -349,7 +349,7 @@ def correct_analog(self, c, data, average_ends=False, dither=False):
 
     @setting(33,
         'Correct Analog FT',
-        data=['*v: Single channel data'],
+        data=['*c: Single channel data'],
```
DanielSank: `'*c[]'` ?

maffoo: I'm inclined to leave this as `*c` for now, which is what it was before and also what we have on the IQ correction settings. If we really wanted to be pedantic about things, then we'd probably want `*c[ns]` for the units, since the fourier transformed data technically has dimensions of time (that's what the envelope code spits out when you evaluate in fourier space, for example), but I don't really want to go there now.

DanielSank: I definitely wouldn't put units on this. The time step should be specified separately and all sequence data should be dimensionless. Anyway, I'm fine to not change the type tag but I can't imagine how it would cause problems to change it to explicitly disallow units.

DanielSank:
LGTM

PR #173, opened 2015-06-01
Modify run_dac_sram call in calibrate to send 20 samples, due to changes in ghz fpga

JulianSKelly:
server.


maffoo:
Could you describe what changes in the server necessitated this?

JulianSKelly:
Right [here](https://github.com/martinisgroup/servers/blob/master/GHzDACs/ghz_fpga_server.py#L1663)

maffoo:
right, thanks. LGTM

PR #174, opened 2015-06-02
Reinitialize BoardGroup's on Direct Ethernet restart.

### LINKED ISSUE #134: ghz_fpga_server fails to reconnect when direct ethernet server is restarted
We don't normally do this, but since we've been trying out the scala version of the direct ethernet server we've been restarting and noticed that after connecting to a board group once, when the direct ethernet server is restarted the ghz_fpga_server does not properly reconnect, so the devices don't appear again. Restarting both servers works, but is annoying.

strj:
Switches the BoardGroups to get context and connection from the
DE server every time before using either. Move the BoardGroup
Init call to aslo reinitialize existing BoardGroups on
DE server reconnecting.

Fixes #134

### DIFF
--- a/GHzDACs/ghz_fpga_server.py
+++ b/GHzDACs/ghz_fpga_server.py
@@ -177,7 +177,7 @@ cmdTime_cycles does not properly estimate sram length
 ### BEGIN NODE INFO
 [info]
 name = GHz FPGAs
-version = 5.0.0
+version = 5.0.1
 description = Talks to DAC and ADC boards
 
 [startup]
@@ -273,8 +273,7 @@ class BoardGroup(object):
         self.fpgaServer = fpgaServer
         self.server = server
         self.port = port
-        self.cxn = server._cxn
-        self.ctx = server.context()
+        self.ctx = None
         #self.sourceMac = getLocalMac(port)
         self.pipeSemaphore = defer.DeferredSemaphore(NUM_PAGES)
         self.pageNums = itertools.cycle(range(NUM_PAGES))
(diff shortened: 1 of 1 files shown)


pomalley:
If I'm reading this correctly, the idea is to no longer get a context in `BoardGroup.__init__` and hang on to it for the life of the BoardGroup, because then when the DE server restarts the context becomes bogus. However, if I recall the way contexts work (correct me if I'm wrong, Matthew), calling `server.context()` creates a new context, which we only want to do once--or rather, once we `connect()` to a DE server we need to keep that context around. So as far as I can tell this would put the wait/load/run packets in a new, different context than that which was created in `__init__` and used for `connect()`, which I don't think we want.

strj:
Hmm, I think you're right: https://github.com/martinisgroup/pylabrad/blob/master/labrad/backend.py#L26
let me fix it up...

maffoo:
Calling `context()` just creates a context identifier, which is a purely client-side notion. It would be ok to create this once and reuse it across restarts of the DE server, as long as you set up the _state_ associated with that context identifier each time the server restarts. But this way works as well and is arguably cleaner. LGTM after we test it on the fridge.

strj:
Great, let me know when it's been tested and if I need to make any modifications or can push to master.

pomalley:
@DanielSank @christopherquintana I think Jules is less busy than Vince at the moment :-) so perhaps you guys can test it there?

This also points to the fact that we need *real* tests for the FPGA server (and all the others...) but that's another issue.

maffoo:
Tested on jules with @christopherquintana, and works great. @strj, please also bump the patch number (https://github.com/martinisgroup/servers/blob/master/GHzDACs/ghz_fpga_server.py#L180). After that, LGTM.

JulianSKelly:
I just want to say that it's incredibly awesome getting your help @strj

strj:
Thanks @JulianSKelly, I'm glad to help.

PR #176, opened 2015-06-03
Convert qubit sequencer config package to scala

maffoo:
See #164


QubitServer/src/main/java/org/labrad/qubits/QubitContext.java
```diff
@@ -21,7 +21,7 @@
 import org.labrad.qubits.channeldata.*;
 import org.labrad.qubits.channels.*;
 import org.labrad.qubits.config.MicrowaveSourceConfig;
-import org.labrad.qubits.config.MicrowaveSourceOffConfig;
+import org.labrad.qubits.config.MicrowaveSourceOffConfig$;
```
pomalley: I gather this dollar sign business is how Java deals with Scala singletons. It seems a bit ugly to me but I guess it's temporary.

QubitServer/src/main/java/org/labrad/qubits/config/AdcAverageConfig.scala
```diff
@@ -0,0 +1,43 @@
...
+    this.criticalPhase = criticalPhase
+  }
+
+  override def interpretPhases(Is: Array[Int], Qs: Array[Int], channel: Int): Array[Boolean] = {
+    require(Is.length == Qs.length, "Is and Qs must have same length!")
+    if (channel != -1) {
+      println("WARNING: interpretPhases for average mode ADC called with demod channel != -1")
+    }
+    (Is zip Qs).map { case (i, q) =>
+      Math.atan2(q, i) < criticalPhase
```
pomalley: not that I think this code ever gets used, but it looks like we've switched the order of the arguments in atan2 here.

QubitServer/src/main/java/org/labrad/qubits/config/AdcBaseConfig.scala
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Map
+
+import org.labrad.data.Request
+
+abstract class AdcBaseConfig(channelName: String, builProperties: Map[String, Long]) {
+
+  /**
+   * number of clock cycles to delay
+   */
+  protected var startDelay: Int = -1
```
pomalley: AFAICT this is different but in fact correct, given, the checks we perform later...

QubitServer/src/main/java/org/labrad/qubits/config/AdcDemodConfig.scala
```diff
@@ -0,0 +1,213 @@
...
+    inUse(channel) = false
+  }
+
+  /**
+   * Set all critical phases.
+   * @param criticalPhases
+   */
+  def setCriticalPhases(criticalPhases: Array[Double]): Unit = {
+    require(criticalPhases.length == MAX_CHANNELS, "Number of critical phases must = number of channels")
+    for (i <- 0 until criticalPhases.length) {
```
pomalley: intellij thinks it's better to say `for (i <- criticalPhases.indices)`...

QubitServer/src/main/java/org/labrad/qubits/QubitContext.java
maffoo: That's right; this and the `MODULE$` business below will both go away when the conversion is done.

QubitServer/src/main/java/org/labrad/qubits/config/AdcDemodConfig.scala
```diff
@@ -0,0 +1,213 @@
...
+      require(criticalPhases(i) >= 0.0 && criticalPhases(i) <= 2*Math.PI,
+          "Critical phases must be between 0 and 2PI");
```
pomalley: semicolon

QubitServer/src/main/java/org/labrad/qubits/config/AdcDemodConfig.scala
```diff
@@ -0,0 +1,213 @@
...
+   * for each channel: ADC Demod Phase
+   *                   ADC Trig Magnitude
+   * @param runRequest The request to which we add the packets.
+   * @author pomalley
+   */
+  def addPackets(runRequest: Request): Unit = {
+    // check that the user has set everything that needs to be set
+    require(startDelay > -1, s"ADC Start Delay not set for channel '$name'")
+    require(stretchLen > -1 && stretchAt > -1, s"ADC Filter Func not set for channel '$name'")
+    require(inUse.exists(_ == true), s"No demod channels activated for channel '$name'")
```
pomalley: intellij says to use contains instead of exists. contains reads a little cleaner to me but mostly because i'm not that familiar with scala yet.

QubitServer/src/main/java/org/labrad/qubits/config/MicrowaveSourceOffConfig.scala
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  override def power: Double = {
+    sys.error("Microwaves are off")
+  }
+
+  override def isOn: Boolean = {
+    false
+  }
+
```
pomalley: I like the whole case object deal, and how it gives us e.g. equals and such.

QubitServer/src/main/java/org/labrad/qubits/config/PreampConfig.scala
```diff
@@ -0,0 +1,61 @@
...
+    val linkName = chName.substring(0, linkNameEnd)
+    val cardId = (chName.substring(linkNameEnd + "Preamp".length() + 2)).toLong
+
+    val data = Data.ofType("(ss)(sw)(s(s(wwww)))(s)")
+    data.get(0).setString("Connect", 0).setString(linkName, 1)
+    data.get(1).setString("Select Card", 0).setWord(cardId, 1)
+    data.get(2).setString("Register", 0).setString(ch.getPreampChannel().toString().toUpperCase(), 1, 0)
+    .setWord(highPass, 1, 1, 0)
+    .setWord(lowPass, 1, 1, 1)
+    .setWord((if (polarity) 1L else 0L), 1, 1, 2)
```
pomalley: intellij thinks the parens around the if-else are unnecessary.

QubitServer/src/main/java/org/labrad/qubits/config/SetupPacket.scala
```diff
@@ -0,0 +1,5 @@
+package org.labrad.qubits.config
+
+import org.labrad.data.Data
+
+class SetupPacket(val state: String, val records: Data)
```
pomalley: heh

pomalley:
Well, again I just said what intellij told me to. ;-)

LGTM pending a fridge test.

QubitServer/src/main/java/org/labrad/qubits/config/AdcAverageConfig.scala
maffoo: Wow, good catch. I think that was actually wrong before; `atan2` wants the y-coordinate first: https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#atan2-double-double-

pomalley: Indeed I figured you wrote it that way because it's correct. ;-) It's (q,i)
in the demod case so let's just make it (q,i) here as well. Honestly I'm
not sure exactly what IQ means for average mode since in average mode as I
understand it you're just getting the time trace.

QubitServer/src/main/java/org/labrad/qubits/config/AdcBaseConfig.scala
maffoo: Actually, I just checked and this is not in fact used anywhere. The `startDelay` value is set directly on the fpga model, and this one never gets set or read (I tried removing the accessor methods and everything still compiles just fine). I'm going to remove these for now since they're not needed.

QubitServer/src/main/java/org/labrad/qubits/config/AdcDemodConfig.scala
maffoo: good tip. thanks, intellij! :)

QubitServer/src/main/java/org/labrad/qubits/config/AdcDemodConfig.scala
maffoo: done

QubitServer/src/main/java/org/labrad/qubits/config/MicrowaveSourceOffConfig.scala
maffoo: yes, case classes and case objects are one of the best features of scala IMO

QubitServer/src/main/java/org/labrad/qubits/config/PreampConfig.scala
maffoo: removed

QubitServer/src/main/java/org/labrad/qubits/config/AdcBaseConfig.scala
maffoo: so, when I take out the member variable itself, things fail to compile because the `reconcile` method checks that it is > -1 or else fails with an error that the delay was not set. But the getter is never called; the only reason this check is passing now is that `startDelay` is getting initialized to the default value of 0, which is obviously greater than -1...

maffoo: #188

pomalley:
Now that I've done this can you review the CZ frequency finder? @DanielSank started but then went away.

maffoo:
Tested on the fridge with @JulianSKelly

PR #178, opened 2015-06-04
Fix bug with wrong index of CYCLE argument processing.

pomalley:
The QS is looking at the wrong index for converting the CYCLE command from number of cycles desired to counter index.

### DIFF
--- a/QubitServer/src/main/java/org/labrad/qubits/jumptable/JumpTable.java
+++ b/QubitServer/src/main/java/org/labrad/qubits/jumptable/JumpTable.java
@@ -12,6 +12,7 @@ import org.labrad.data.Request;
  * Basic logic for the jump table.
  */
 public class JumpTable {
+  private static final int NUM_COUNTERS = 4; // TODO: get num counters from hardware
   private final List<String> entryNames;
   private final List<Data> entryArguments;
   private long[] counters;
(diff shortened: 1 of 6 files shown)


QubitServer/src/main/java/org/labrad/qubits/jumptable/JumpTable.java
```diff
@@ -38,11 +38,11 @@ public void addEntry(String name, Data argument) {
     if (name.equals("CYCLE")) {
       List<Data> args = argument.getClusterAsList();
       Preconditions.checkArgument(args.size() == 4, "Cycle must have 4 arguments; currently has " + args.toString());
-      if (countersUsed == 3) {     // TODO: get num counters from hardware
+      if (countersUsed == 4) {     // TODO: get num counters from hardware
```
maffoo: can we factor this out into a class-level static final constant for now? The TODO about getting the value from hardware instead can go with the constant.

pomalley:
Did that

pomalley:
Julian says it works.

pomalley:
Also what's the deployment strategy these days?

maffoo:
Please move the TODO comment up to the constant as well.

As for deployment, you'll want to bump the version on the server, then just run `sbt packArchive` to produce a `.tar.gz` file in the `target` directory of the project. Unpack that on the server under `C:\labrad\servers` and delete the version that was there. Refresh the node and you should see it pick up the new version. Automation of all this coming soon...

pomalley:
OK. Just wanted to make sure we're no longer committing binaries.

Found another bug to do with mutable data, and added an offset to deal with JT entry index counting on the fpga_server side of things. This has now been tested to work on Vince, and Julian should have a pyle PR incoming soon that has fixes on that end.

maffoo:
LGTM

pomalley:
Sweet, thanks.

PR #181, opened 2015-06-08
Consolidate code shared by data_vault and data_vault_multihead

maffoo:
We move the code for these servers into the datavault package and make the server class for the multi-headed data vault into a subclass of the single-headed server class, so that the two can share the setting code. The server launch scripts are reduced basically just to shims that load some settings and then launch the server itself.

Depends on martinisgroup/pylabrad#114

### DIFF
--- a/data_vault.py
+++ b/data_vault.py
@@ -30,410 +30,75 @@ timeout = 5
 ### END NODE INFO
 """
 
-# CHANGELOG
-#
-# 31 January 2012 - Dan Sank and Ted White
-# Added setting to dump existing open sessions in Session._sessions.
-# We did this because the data vault seems to be leaking memory.
-# We think sessions are never getting garbage collected
-# even after all listening contexts expire. We want to be able to view all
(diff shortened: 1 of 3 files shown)


data_vault.py
```diff
@@ -30,410 +30,73 @@
...
+    yield reg.cd(path, True)
+    (dirs, keys) = yield reg.dir()
+    if nodename in keys:
+        datadir = yield reg.get(nodename)
+    elif '__default__' in keys:
+        datadir = yield reg.get('__default__')
+    else:
+        print 'Could not load repository location from registry.'
+        print 'Please enter data storage directory or hit enter to use the current directory:'
+        datadir = os.path.expanduser(raw_input('>>>'))
```
ejeffrey: Print CWD here so that the user can see what hitting enter does

ejeffrey:
LGTM.  I had some other comments, but since it was mostly all reorganized code rather than changed, it should be kept separate from this.

Issue #109, opened 2015-04-20
Reorganize code in servers repo


maffoo:
The servers repo is a mishmash of servers, scripts and library code, without a very coherent organization. This makes it hard to know how to get it on your python path, and also makes it hard to know inside code itself how to import and use other code.

From the discussion of #108:

ghzdac is a package containing library code for the deconvolution; this code should be importable from elsewhere, so it should have an __init__.py and it should be on the python path.

ghzdac_cal contains runnable scripts for doing calibration; these scripts should not be in a python package, given python's separation of scripts from library code, and so this should not have an __init__.py file. Or at the very least, the scripts should be refactored to put all the logic in a python package, and just have runnable stubs in a separate scripts directory (which is not a python package) that just imports and runs the appropriate library code. The stubs could also do some path manipulation to make sure that the corresponding package directories from the same git checkout are on the python path; this is a concession we've discussed for people who are on windows and want scripts to be double-clickable in a gui environment on a shared machine that may have different people's personal checkouts and so the system python path may point at the wrong location.

Finally, GHzDACs contains server code as well as other miscellanea (looks like there are some scripts in there, etc.) Again, anything that should be top-level runnable should be refactored into library code and a script stub in a non-package directory. I think the server itself should also receive the same refactoring treatment, since it needs to be top-level runnable by the node.

So, roughly speaking, I propose the following structure for the servers repo:

ghzdac/
    __init__.py
    adc.py
    calibrate.py
    ...

scripts/
    GHz_DAC_bringup.py
    ...

servers/
    ghz_fpga_server.py
    ...
Note that we would remove the top-level __init__.py file in the servers repo; this whole thing is not a python package. Also note that the scripts and servers folders do not have __init__.py files; these are also not python packages. The servers checkout would no longer have to be put inside an extra directory when checked out and the checkout folder itself could be added to the pythonpath (at least on non-shared machines or in virtualenvs) and this would make the code in the ghzdac package available. Also note that the ghzdac package as I've sketched it contains both the calibration and server code mashed together; we'd probably want to structure that differently, but I think having a clean separation between scripts and library code (+ tools like pycharm) makes that easier.

ejeffrey:
Is it valuable to be able to import the actual server code?  We could put
99% of the server code in the package directory, and then just make stubs
with the node info block and calls to util.runServer().  I don't know if
this would actually get us anything useful.

Evan

DanielSank:
[...]

PR #187, opened 2015-06-09
Add support for storing data in hdf5 format

maffoo:
This is @ejeffrey's work, I've just done some cleanup of the branch history and rebased off current master. Let's fill out the tests before merging this in.


datavault/__init__.py
```diff
@@ -216,26 +217,30 @@ def exclude(entries, tag, tags):
...
 
     def listDatasets(self):
         """Get a list of dataset names in this directory."""
         files = os.listdir(self.dir)
-        files.sort()
-        return [filename_decode(s[:-4]) for s in files if s.endswith('.csv') or s.endswith('.bin')]
-
-    def newDataset(self, title, independents, dependents):
+        filenames = []
+        for s in files:
```
JulianSKelly: We no longer support .bin files? What were those?

datavault/__init__.py
```diff
@@ -216,26 +217,30 @@ def exclude(entries, tag, tags):
...
+            base, _, ext = s.rpartition('.')
+            if ext in ['csv', 'hdf5']:
```
JulianSKelly: We no longer support .bin files? What were those?

datavault/__init__.py
```diff
@@ -257,7 +262,7 @@ def openDataset(self, name):
 
         filename = filename_encode(name)
         file_base = os.path.join(self.dir, filename)
-        if not (os.path.exists(file_base + '.csv') or os.path.exists(file_base + '.bin')):
+        if not (os.path.exists(file_base + '.csv') or os.path.exists(file_base + '.bin') or os.path.exists(file_base + '.hdf5')):
```
JulianSKelly: Why .bin here and not above?

datavault/__init__.py
```diff
@@ -318,158 +323,78 @@ def getTags(self, sessions, datasets):
...
+    """
+    This object basically takes care of listeners and notifications.
+    All the actual data or metadata access is proxied through to a
+    backend object.
+    """
+    def __init__(self, session, name, title=None, num=None, create=False, independents=[], dependents=[], extended=False):
         self.hub = session.hub
         self.name = name
         file_base = os.path.join(session.dir, filename_encode(name))
-        self.infofile = file_base + '.ini'
```
JulianSKelly: What's the idea behind getting rid of the association with the ini file? Is it never used?

datavault/__init__.py
maffoo: `.bin` was an experiment that happened on a branch that this branch was derived from. We decided to go for hdf5 instead. I think this must have made it into master by mistake when we merged some other features from that branch. We've never actually used a version of the data vault that wrote `.bin` files.

datavault/__init__.py
maffoo: It should go away; will remove.

datavault/__init__.py
```diff
@@ -318,158 +323,78 @@ def getTags(self, sessions, datasets):
         return sessTags, dataTags
 
 class Dataset(object):
-    def __init__(self, session, name, title=None, num=None, create=False, independents=[], dependents=[]):
+    """
+    This object basically takes care of listeners and notifications.
+    All the actual data or metadata access is proxied through to a
+    backend object.
+    """
+    def __init__(self, session, name, title=None, num=None, create=False, independents=[], dependents=[], extended=False):
```
JulianSKelly: num has no function

datavault/__init__.py
maffoo: the file management is moved into the backend. The CSV backend which is used to read old files still uses a separate .ini file. But the HDF5 backends that store newer datasets also put the metadata inside the hdf5 file, so everything is self-contained and there's no need for a separate .ini file.

datavault/__init__.py
```diff
@@ -216,26 +217,30 @@ def exclude(entries, tag, tags):
...
         num = self.counter
         self.counter += 1
         self.modified = datetime.now()
 
         name = '%05d - %s' % (num, title)
-        dataset = Dataset(self, name, title, create=True, independents=independents, dependents=dependents)
+        dataset = Dataset(self, name, title, create=True,
+                          independents=independents,
+                          dependents=dependents,
+                          extended=extended)
```
JulianSKelly: What is extended exactly?

datavault/backend.py
```diff
@@ -216,7 +439,378 @@ def hasMore(self, pos):
...
+            raw_comments = self.dataset.attrs['Comments'][start:]
+        else:
+            raw_comments = self.dataset.attrs['Comments'][start:start+limit]
+        comments = [(datetime.datetime.fromtimestamp(c[0]), str(c[1]), str(c[2])) for c in raw_comments]
+        return comments, start+len(comments)
+
+    def numComments(self):
+        return len(self.dataset.attrs['Comments'])
+
+class ExtendedHDF5Data(HDF5MetaData):
```
JulianSKelly: Oh, this is extended.

datavault/backend.py
```diff
@@ -216,7 +439,378 @@ def hasMore(self, pos):
...
+            struct_data = self.dataset[start:start+limit]
+        return struct_data, start + struct_data.shape[0]
+
+    def __len__(self):
+        return self.dataset.shape[0]
+
+    def hasMore(self, pos):
+        return pos < len(self)
+
+class SimpleHDF5Data(HDF5MetaData):
```
JulianSKelly: Is this not just a special case of HDF5MetaData? Why do we need a whole class? Is it just for backwards compatibility?

datavault/backend.py
```diff
@@ -216,7 +439,378 @@ def hasMore(self, pos):
...
+        return self.dataset.shape[0]
+
+    def hasMore(self, pos):
+        return pos < len(self)
+
+def open_hdf5_file(filename):
+    """Factory for HDF5 files.  
+
+    We check the version of the file to construct the proper class.  Currently, only two
+    options exist: version 2.0.0 -> legacy format, 3.0.0 -> extended format.
```
JulianSKelly: So version 1 is data taken before this version of the data vault
version 2 is the old 2d array data sets stored in hdf5
and version 3 is the fullhdf5?

datavault/__init__.py
maffoo: good catch. removed.

JulianSKelly:
This looks good to me from a high level, but it's hard for me to check specifics.

datavault/__init__.py
JulianSKelly: Oh that's nice

datavault/backend.py
maffoo: The idea is that the metadata base class deals with just the metadata part, which is common between the simple and extended formats. It's really an abstract class. The subclasses for simple and extended add the logic for storing and loading the data itself. It would certainly be possible to restructure things.

datavault/backend.py
maffoo: yup

ejeffrey: Yes.  If you use the old call to datavault.new(), you will generate a
version 2.0 dataset, that only supports 2D arrays, and is API compatible
with the old functions.  When you call new_ex() you generate a version 3.0
dataset that supports all the new features but may not work with the old
add() and get() calls.  I have now made it so that if you use the new API
to create a version 3 dataset that happens to be a 2-D array of floats, it
will still work with the old API, so maybe this isn't needed any more.

datavault/backend.py
ejeffrey: The idea is the metadata base class handles all the parameters and stuff,
the subclass handles the data access.  Technically it would be better to
use composition here, but since they are stored in the same underlying data
store (the HDF5 file) which they have to cooperatively access, that looked
kind of tricky.

datavault/backend.py
```diff
@@ -216,7 +439,378 @@ def hasMore(self, pos):
...
+            else:
+                raise RuntimeError("Invalid type tag {}".format(ttag))
+
+        self.file.create_dataset('DataVault', (0,), dtype=dtype, maxshape=(None,))
+        HDF5MetaData.initialize_info(self, title, indep, dep)
+
+    @property
+    def file(self):
+        return self._file()
+    @property
```
strj: missing a blank line above this one

datavault/backend.py
```diff
@@ -99,15 +320,19 @@ def _saveData(self, data):
         f.flush()
 
     def addData(self, data):
+        if transpose:
```
strj: this will:
NameError: name 'transpose' is not defined

datavault/backend.py
```diff
@@ -99,15 +320,19 @@ def _saveData(self, data):
...
+            raise RuntimeError("Transpose specified for simple data format: not supported")
         if not len(data) or not isinstance(data[0], list):
             data = [data]
         if len(data[0]) != self.cols:
```
strj: self.cols is not defined either

datavault/backend.py
```diff
@@ -216,7 +439,379 @@ def hasMore(self, pos):
...
+        """Load and save do nothing because HDF5 metadata is accessed live"""
+        pass
+
+    def save(self):
+        """Load and save do nothing because HDF5 metadata is accessed live"""
+        pass
+
+    @property
+    def dtype(self):
+        return self.dataset.dtype
```
strj: maybe i'm missing something but I can't see where self.dataset get's assigned?

datavault/backend.py
maffoo: good catch; will remove.

datavault/backend.py
maffoo: this gets set in the `IniData` base class, either when the data is loaded for an existing dataset (https://github.com/martinisgroup/servers/blob/u/maffoo/hdf5_data/datavault/backend.py#L132) or in the `initialize_info` method for new datasets (https://github.com/martinisgroup/servers/blob/u/maffoo/hdf5_data/datavault/backend.py#L210)

datavault/backend.py
```diff
@@ -216,7 +439,379 @@ def hasMore(self, pos):
...
+
+    def getParameter(self, name, case_sensitive=True):
+        """Get a parameter from the dataset."""
+        keyname = 'Param.{}'.format(name)
+        if case_sensitive:
+            if keyname in self.dataset.attrs:
+                return labrad_urldecode(self.dataset.attrs[keyname])
+        else:
+            for k in self.dataset.attrs:
+                if k.lower == keyname.lower:
```
strj: k.lower -> k.lower()
keyname.lower -> keyname.lower()

datavault/backend.py
```diff
@@ -51,24 +91,205 @@ def size(self):
...
+        return type_tag
+
+    def getTransposeType(self):
+        units = []
+        for var in self.independents + self.dependents:
+            units.append('*v[{}]'.format(var.unit))
+        type_tag = '({})'.format(','.join(units))
+        return type_tag
+
+    def addParam(self, name, data):
```
strj: This does not do any checking of data for labrad types whereas HDF5MetaData does. This could cause a breakage if some weird data type was being stored using addParam.

datavault/backend.py
```diff
@@ -51,24 +91,205 @@ def size(self):
...
+                    return p['data']
+            else:
+                if p['label'].lower() == name.lower():
+                    return p['data']
+        raise errors.BadParameterError(name)
+
+    def getParamNames(self):
+        return [p['label'] for p in self.parameters]
+
+    def addComment(self, user, comment):
```
strj: Note that this stores comments without converting them to strings (different to HDF5MetaData). a save-load cycle will convert it to strings though. Don't think it's important, just wanted to highlight the differences

datavault/backend.py
strj: ah gotcha, I missed that.

datavault/backend.py
```diff
@@ -179,26 +403,25 @@ def _saveData(self, data):
         f.flush()
 
     def addData(self, data):
```
strj: The parent addData accepts a single tuple and converts it to a list (l325), this one doesn't - is that intentional?

datavault/backend.py
```diff
@@ -216,7 +439,379 @@ def hasMore(self, pos):
...
+
+    @property
+    def file(self):
+        return self._file()
+
+    @property
+    def dataset(self):
+        return self.file["DataVault"]
+
+    def addData(self, data):
```
strj: CsvListData and CsvNumpyData do validation on the number of rows and this version does not - is that intended?

datavault/backend.py
strj: ExtendedHDF5Data and SimpleHDF5Data don't have a .data attribute whereas CsvListData and CsvNumpyData do - is that intentional?

datavault/backend.py
```diff
@@ -216,7 +439,379 @@ def hasMore(self, pos):
...
+        """Adds one or more rows or data from a 2D array of floats."""
+        new_rows = data.shape[0]
```
strj: This implies that data has to be a numpy.ndarray, the other addData functions aren't as restrictive, is this intended?

datavault/backend.py
```diff
@@ -216,7 +439,379 @@ def hasMore(self, pos):
...
+                shape_tag = '*2'
+                comment = util.braced('N,' + str(col.shape[0]))
+            else:
+                shape_tag = '*'
+                comment = ''
+            column_type.append(shape_tag + base_type + unit_tag + comment)
+        type_tag = '({})'.format(','.join(column_type))
+        return type_tag
+
+    def addParam(self, name, data):
```
strj: IniData also checks for parameter name collisions and this version doesn't - is that intended?

strj:
I've got a whole bunch of unit tests written for datavault/backend.py and datavault/util.py, have a look at 
https://github.com/martinisgroup/servers/blob/add-datavault-hdf5-tests/datavault/backend_test.py

datavault/backend.py
maffoo: fixed

datavault/backend.py
maffoo: I think the number of columns will get checked by the underlying hdf5 library, since we're appending to a typed array. We should have tests for this, though.

datavault/backend.py
maffoo: yeah, we've been lax here because this only gets called from the server frontend, and we're relying on that to pass only valid labrad data types through. We should have stronger checks to make this more robust.

datavault/backend.py
maffoo: similar to the addParam case, we are relying on the labrad server frontend to pass appropriate types in here. But you're right that we should make this robust on its own.

datavault/backend.py
maffoo: all the addData functions should assume that they'll be passed a numpy array. That wasn't true in the past; we originally wanted this server to work with or without numpy, but at some point decided it's not onerous to require numpy and makes things much faster, but we likely did not update the code to reflect the use of numpy arrays everywhere.

datavault/backend.py
maffoo: That was not intended. I'll add the check.

maffoo:
@strj, those tests look great. Just a couple of comments: the organization that we've adopted in pyle, which is where we've been writing the most python tests, is to put tests for a given package into a `test` subpackage and name the files like `test_foo.py`. If you could reorganize your code accordingly, I'd be happy to merge it into this branch, e.g. `datavault/test/test_backend.py` and `datavault/test/test_util.py` (along with an empty `datavault/test/__init__.py` to make it a proper package). In addition, the overall package structure of this repo is a bit in flux; I don't like the idea of cloning this into, say, the `servers` directory and then importing things as `servers.datavault` because that requires adding the directory _into which this was cloned_ to the python path, instead of the cloned directory itself. You'll note that `data_vault.py` and `data_vault_multihead.py` import code from the `datavault` package, which assumes the latter convention, that the cloned directory is on the python path, not its parent. Anyway, until we sort this out, perhaps the best thing to do would be to use relative imports in the test code, e.g.:
```python
from .. import backend, errors
```
If you have thoughts on code organization, please comment on #109.

strj:
@maffoo Hmm, relative imports require hackery like https://github.com/arruda/relative_import_example/blob/master/my_package/my_pyscript.py in each test script you want to run which I think is best to avoid if possible.

How about keeping with making each server a package then doing
```python
from datavault import backend, errors
```
This keeps the requirement of adding the repo to the path but no more than that at least.

I'm not sure if #109 is the right place to mention it (overall I agree with the organization proposed there) but having foo.py & foo_test.py right next to each other helps remind folks (well me at least) that there are tests you need to think about. And _not_ seeing foo_test.py when there is foo.py, bar.py, and bar_test.py in the directory helps to push the idea that there really ought to be a foo_test.py.

I've pushed an update to my branch so feel free to merge it in here.

strj:
I've pushed a bunch more tests in https://github.com/martinisgroup/servers/blob/add-datavault-hdf5-tests/datavault/test/

py.test with coverage shows:
```
Name                 Stmts   Miss  Cover
----------------------------------------
datavault/__init__     295     17    94%
datavault/backend      561     31    94%
datavault/errors        35      3    91%
datavault/server       302    106    65%
datavault/util          21      4    81%
----------------------------------------
TOTAL                 1214    161    87%
```
I'm still going to flesh out the datavault/server tests to get to above 90%.

ejeffrey:
Thanks, this is really great.

maffoo:
I've done some branch surgery to rebase this off master and add the unit tests from @strj. We'll see how the new tests do in teamcity.

strj:
Looks like teamcity didn't have any problems with the tests. I've the last bunch for server.py to https://github.com/martinisgroup/servers/blob/add-datavault-hdf5-tests/datavault/test/

Final coverage tally:
```
Name                 Stmts   Miss  Cover
----------------------------------------
datavault/__init__     295     13    96%
datavault/backend      561     31    94%
datavault/errors        35      3    91%
datavault/server       302     16    95%
datavault/util          21      4    81%
----------------------------------------
TOTAL                 1214     67    94%
```

strj:
@maffoo @ejeffrey I just realized there's one thing that's apparent from the server.py code that I didn't test: switching context's. I.e. adding a bunch of data / things with context_foo and then calling get with context_bar, checking that there's no 'leaking' of data. Should I add tests for that too?

ejeffrey:
This is deployed and working now, including data taking with pyle, the grapher, and the datavault wrapper.  I plan on letting it run for a couple more days and then merging.

The only thing that is a little concerning is that there is no fallback if you don't have libhdf5 or h5py installed.  I think that is OK, handling that is too tricky, and you wouldn't want to "fall back" to CSV when you already have HDF5 datasets.

ejeffrey:
Reminder: bump the version before merging.

maffoo:
LGTM. Woot!!

datavault/__init__.py
```diff
@@ -198,9 +198,10 @@ def access(self):
     def listContents(self, tagFilters):
         """Get a list of directory names in this directory."""
         files = os.listdir(self.dir)
-        files.sort()
         dirs = [filename_decode(s[:-4]) for s in files if s.endswith('.dir')]
-        datasets = [filename_decode(s[:-4]) for s in files if s.endswith('.csv') or s.endswith('.bin')]
+        csv_datasets = [filename_decode(s[:-4]) for s in files if s.endswith('.ini') and s.lower() != 'session.ini' ]
```
DanielSank: I'd parametrize the 4 here. Also the 5 on the next line.

datavault/__init__.py
DanielSank: The `if` statement with no `else` reads strangely to me. If the idea is to ignore everything that's not csv or hdf5, I would explain that in a comment. I would also make a `SUPPORTED_FILETYPES` variable instead of the inline code.

datavault/__init__.py
```diff
@@ -216,26 +217,30 @@ def exclude(entries, tag, tags):
...
                 filter = include
-            #print filter.__name__ + ':', tag
-            #print 'before:', dirs, datasets
             dirs = filter(dirs, tag, self.session_tags)
             datasets = filter(datasets, tag, self.dataset_tags)
-            #print 'after:', dirs, datasets
         return dirs, datasets
 
     def listDatasets(self):
         """Get a list of dataset names in this directory."""
```
DanielSank: I would mention that the list is sorted (alphabetically).

datavault/__init__.py
```diff
@@ -216,26 +217,30 @@ def exclude(entries, tag, tags):
...
-
-    def newDataset(self, title, independents, dependents):
+        filenames = []
+        for s in files:
+            base, _, ext = s.rpartition('.')
+            if ext in ['csv', 'hdf5']:
+                filenames.append(filename_decode(base))
+        return sorted(filenames)
+
+    def newDataset(self, title, independents, dependents, extended=False):
```
DanielSank: Doc string?

datavault/__init__.py
```diff
@@ -257,7 +262,7 @@ def openDataset(self, name):
 
         filename = filename_encode(name)
         file_base = os.path.join(self.dir, filename)
-        if not (os.path.exists(file_base + '.csv') or os.path.exists(file_base + '.bin')):
+        if not (os.path.exists(file_base + '.csv') or os.path.exists(file_base + '.hdf5')):
```
DanielSank: Parameterizing the supported file extensions would clean this up.

DanielSank:
Uh, ok... maybe consider my comments as an issue filed on this code? :P

datavault/__init__.py
```diff
@@ -318,158 +323,78 @@ def getTags(self, sessions, datasets):
         return sessTags, dataTags
 
 class Dataset(object):
-    def __init__(self, session, name, title=None, num=None, create=False, independents=[], dependents=[]):
+    """
+    This object basically takes care of listeners and notifications.
+    All the actual data or metadata access is proxied through to a
+    backend object.
+    """
+    def __init__(self, session, name, title=None, create=False, independents=[], dependents=[], extended=False):
```
DanielSank: Line too long.

datavault/__init__.py
```diff
@@ -318,158 +323,78 @@ def getTags(self, sessions, datasets):
         return sessTags, dataTags
 
 class Dataset(object):
-    def __init__(self, session, name, title=None, num=None, create=False, independents=[], dependents=[]):
+    """
+    This object basically takes care of listeners and notifications.
+    All the actual data or metadata access is proxied through to a
+    backend object.
+    """
```
DanielSank: It's helpful (and part of the style guide) to list the important attributes of a class.

datavault/__init__.py
```diff
@@ -318,158 +323,78 @@ def getTags(self, sessions, datasets):
...
+        return '.'.join(str(x) for x in v)
 
     def access(self):
         """Update time of last access for this dataset."""
-        self.accessed = datetime.now()
+        self.data.access()
         self.save()
 
-    def makeIndependent(self, label):
+    def makeIndependent(self, label, extended):
```
DanielSank: It's not obvious what `extended` is.

datavault/__init__.py
```diff
@@ -547,14 +461,10 @@ def addComment(self, user, comment):
...
 
     def getComments(self, limit, start):
-        if limit is None:
-            comments = self.comments[start:]
-        else:
-            comments = self.comments[start:start+limit]
-        return comments, start + len(comments)
+        return self.data.getComments(limit, start)
 
     def keepStreamingComments(self, context, pos):
```
DanielSank: Could use a docstring. It's not obvious to me what this does.

datavault/backend.py
```diff
@@ -10,30 +17,67 @@
...
 DATA_TIMEOUT = 300 # how long to keep data in memory if not accessed
+DATA_URL_PREFIX = 'data:application/labrad;base64,'
+
+def time_to_str(t):
+    return t.strftime(TIME_FORMAT)
+
+def time_from_str(s):
+    return datetime.datetime.strptime(s, TIME_FORMAT)
+
+def labrad_urlencode(data):
```
DanielSank: Some specification of the url in a doc string would be nice.

datavault/backend.py
```diff
@@ -49,26 +93,215 @@ def size(self):
...
+            label = S.get(sec, 'Label', raw=True)
+            units = S.get(sec, 'Units', raw=True)
+            categ = S.get(sec, 'Category', raw=True)
+            return Dependent(label=categ, legend=label, shape=(1,), datatype='v', unit=units)
+        count = S.getint(gen, 'Dependent')
+        self.dependents = [getDep(i) for i in range(count)]
+
+        self.cols = len(self.independents + self.dependents)
+
+        def getPar(i):
```
DanielSank: If `Par` is "parameter" I'd spell that out.

datavault/backend.py
```diff
@@ -102,12 +336,14 @@ def addData(self, data):
...
-            raise BadDataError(self.cols, len(data[0]))
+            raise errors.BadDataError(self.cols, len(data[0]))
 
         # append the data to the file
         self._saveData(data)
 
-    def getData(self, limit, start):
+    def getData(self, limit, start, transpose, simpleOnly):
+        if transpose:
+            raise RuntimeError("Transpose specified for simple data format: not supported")
```
DanielSank: Would reword this to `"Cannot transpose CSV data set {}".format(self.filename)`

datavault/backend.py
```diff
@@ -49,26 +93,215 @@ def size(self):
...
+        self.infofile = filename[:-4] + '.ini'
+        self.reactor = reactor
 
     @property
     def file(self):
         return self._file()
 
     @property
+    def version(self):
+        return np.asarray([1,0,0], np.int32)
```
DanielSank: Should this be a class variable?

PR #189, opened 2015-06-12
Fix the failing ghz_fpga_test.

### LINKED ISSUE #162: ghz_fpga_server tests failing
The tests are being picked up and run as part of the data_vault server tests. A recent failure is here: https://matrix-reloaded.physics.ucsb.edu/teamcity/viewLog.html?buildId=1398&tab=buildResultsDiv&buildTypeId=Servers_DataVault

strj:
It seems that the test was written assuming an IDLE time offset
of -1, but in GHzDACs.Cleanup.dac.DAC_Build15 there is:
JT_IDLE_OFFSET = -0

Fixes #162.

### DIFF
--- a/GHzDACs/test/test_fpga_server.py
+++ b/GHzDACs/test/test_fpga_server.py
@@ -28,6 +28,9 @@ from labrad.units import Value
 NUM_DACS = 3
 DAC_BUILD = 15
 
+_JUMP_TABLE_IDLE_OFFSET = 0
+_JUMP_TABLE_FROM_ADDR_OFFSET = -2
+_JUMP_TABLE_END_ADDR_OFFSET = -3
 
 class TestFPGAServer(object):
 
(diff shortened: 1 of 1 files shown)


strj:
I'm not sure if the offset ought to be -1 or 0... presumably it's the test that's wrong.

maffoo:
@pomalley, you changed this indexing recently, if I'm not mistaken. Can you take a look at @strj's fix? I'm going to take a look at creating some test jobs for the servers repo because we want to catch these test breakages early.

pomalley:
This is the correct fix. The absolutely proper thing to do (I guess?) would be to incorporate the offsets into the test, i.e. use the class level JT_IDLE_OFFSET, etc.

strj:
I generally lean towards not including constants from the objects being tested in the tests directly since it prevents refactorings that accidentally change the the constant and it's fairly easy to update. I'd rather add duplicate constants in the test file to make it easier to understand.

pomalley:
I'm OK with that. LGTM (we can skip the fridge test b/c the change is only
to a test (famous last words))

strj:
:) cool, I'll go ahead with the merge then

PR #191, opened 2015-06-19
Convert mem, resources, and templates code to scala

maffoo:
See #164 

I'm including changes to three packages in this PR because I think we're getting the hang of this, and because these classes are particularly boilerplate-y and not very interesting. As usual, I've separated the file rename changes from the code changes, so for the clearest diffs, please look at the code change diffs separately: cf6ecd8, 036a976, and 165b810.


pomalley:
OK @maffoo, I think it's basically good. Time to test it out.

maffoo:
tested on fridge with @pomalley

PR #193, opened 2015-06-25
Convert controller, jumptable, proxies and util files to scala

maffoo:
See #164


pomalley:
Basically LGTM, given the usual.

PR #197, opened 2015-06-30
Add grace period to cryo notifier

### LINKED ISSUE #196: Cryo notifier is super annoying
The Cryo Notifier is super annoying.  Whenever the manager or lakeshore server is restarted, it will send out an alert because there is no valid temperature data.  This causes people to assume that the alerts are meaningless because they happen so often.  I have modified the notifier to not send out alerts unless there is an exception that lasts more than 2 minutes (grace period configurable via registry).  Available in u/ejeffrey/cryo_notifier_deannoyify .  If someone tests it and it doesn't break horribly I will make a PR.

ejeffrey:
Fixes #196 by making the cryo notifier wait a short period before sending notifications.  This should allow the lakeshore to start up and complete a scan of the thermometers without sending any bogus warnings.

### DIFF
--- a/cryo_notifier.py
+++ b/cryo_notifier.py
@@ -166,6 +166,7 @@ class CryoNotifier(LabradServer):
         else:
             self.node = 'node_'+self.path[0]
         print self.path
+        self.last_ok_time = time.time()
         yield start_server(self.client, self.node, 'Telecomm Server')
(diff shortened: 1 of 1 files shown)


DanielSank:
Code looks fine. Should be test driven.

PR #200, opened 2015-07-14
FPGA: Fill out docstrings for LVDS,FIFO Bringup

jwenner:
Fill out docstrings for functions used in DAC LVDS and FIFO bringup.

### DIFF
--- a/GHzDACs/Cleanup/dac.py
+++ b/GHzDACs/Cleanup/dac.py
@@ -677,8 +677,19 @@ class DAC_Build7(DAC):
 
     @inlineCallbacks
     def _checkPHOF(self, op, fifoReadings, counterValue):
-        # Determine for which PHOFs (FIFO offsets) the FIFO counter equals
-        # counterValue.
+        """Determine correct PHOF for FIFO counter of counterValue.
+
+        If we find a successful PHOF (phase offset), set PHOF and read FIFO
+        counter to verify that FIFO counter equals counterValue.
+
+        If no PHOF gives an acceptable FIFO counter, return (-1,False).
+
+        :param int op: 2 for DAC A, 3 for DAC B
+        :param list[int] fifoReadings: FIFO counters for each PHOF
+        :param int counterValue: Desired FIFO counter.
+        :return: (PHOF for given counterValue, success)
+        :rtype: (int, bool)
+        """
         # Relying on indices matching PHOF values.
         PHOFS = np.where(fifoReadings == counterValue)[0]
         # If no PHOF can be found with the target FIFO counter value...
@@ -765,21 +776,31 @@ class DAC_Build7(DAC):
         return self.testMode(self._setPolarity, chan, invert)
 
     def setLVDS(self, cmd, sd, optimizeSD):
-        """
-        INPUTS
-        cmd - int: 2 for DAC A, 3 for DAC B
-        sd - :
-        optimizeSD - bool:
-        
-        Returns (success, MSD, MHD, t, (range(16), MSDbits, MHDbits), checkHex)
-        success - bool: 
-        MSD - int:
-        MHD - int:
-        t - :
-        range(16)
-        MSDbits - :
-        MHDbits - :
-        checkHes - :
+        """Calibrate DAC LVDS.
+        Align DAC clocks for LVDS, varying SD (sample delay).
+        :param int cmd: 2 for DAC A, 3 for DAC B
+        :param int/None sd: If int, SD value to be set. If None, SD set to value
+                            in registry. Ignored if optimizeSD=True.
+        :param bool optimizeSD: Whether to follow procedure from data sheet to
+                                determine SD
+        :return: success - If LVDS bringup successful. MSD and MHD should only
+                           flip once with flip locations within one bit of each
+                           other. If varies, could mean clock noise.
+                 MSD - Measured sample delay. If optimizeSD, where MSD flips
+                       when MHD=SD=0. Else -1.
+                 MHD - Measured hold delay. If optimizeSD, where MHD flips when
+                       MSD=SD=0. Else -1.
+                 t - SD value set
+                 tuple - Transpose of [SD, MSD(SD), MHD(SD) for SD in range(16)]
+                 checkHex - In binary, '0bABC', where
+                            A=1: An LVDS input was above the IEEE input voltage
+                                 specification limit
+                            B=1: An LVDS input was below the IEEE input voltage
+                                 specification limit
+                            C=1: Sampling in correct data cycle
+        :rtype: (bool, int, int, int,
+                 tuple(list[int], list[bool], list[bool]),
+                 int)
         """
 
         @inlineCallbacks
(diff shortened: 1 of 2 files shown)


GHzDACs/Cleanup/dac.py
```diff
@@ -677,8 +677,25 @@ def _setPolarity(self, chan, invert):
 
     @inlineCallbacks
     def _checkPHOF(self, op, fifoReadings, counterValue):
-        # Determine for which PHOFs (FIFO offsets) the FIFO counter equals
-        # counterValue.
+        """ Determine correct PHOF for FIFO counter of counterValue
```
pomalley: To be completely 100% correct, I believe the standard is to not have a after the `"""` and to end this line with a period.

GHzDACs/Cleanup/dac.py
```diff
@@ -677,8 +677,25 @@ def _setPolarity(self, chan, invert):
...
+
+        If finds such a PHOF, sets PHOF and reads back FIFO counter to
+        verify that FIFO counter equals counterValue.
```
pomalley: "If finds..."? Is this supposed to be "If we find a valid PHOF, set PHOF and..."

GHzDACs/Cleanup/dac.py
```diff
@@ -677,8 +677,25 @@ def _setPolarity(self, chan, invert):
...
+
+        If no PHOF can be found to get an acceptable FIFO counter, PHOF will
+        be returned as -1, and success will be False.
```
pomalley: Generally try to keep things in the imperative. I suggest: "If no PHOF can be found to get an acceptable FIFO counter, return (-1, False)."

GHzDACs/Cleanup/dac.py
```diff
@@ -677,8 +677,25 @@ def _setPolarity(self, chan, invert):
...
+        :param list[int] fifoReadings: FIFO counters for each PHOF
+                                       (FIFO phase offset)
+        :param int counterValue: Desired FIFO counter.
+        :return: PHOF - PHOF (phase offset) required for targetFifo.
+                        If not successful, -1.
+                 found - If procedure successful. Requires PHOF exists
+                         for FIFO counter to be counterValue and, after
+                         setting PHOF, that FIFO counter is counterValue
+        :rtype: (int, bool)
+        """
```
pomalley: I think it would be better to define PHOF (as phase offset? FIFO phase offset?) when it's first introduced, after the first line. Since we already know that the goal is to find the valid PHOF, I would condense the :return: part to:
```
:return: (PHOF for given counterValue, success)
```

GHzDACs/Cleanup/dac.py
```diff
@@ -765,21 +782,33 @@ def setPolarity(self, chan, invert):
...
+                       flips when MHD=SD=0. Else -1.
+                 MHD - Measured hold delay. If optimizeSD, where MHD
+                       flips when MSD=SD=0. Else -1.
+                 t - SD value set
+                 tuple - Transpose of [SD, MSD(SD), MHD(SD) for SD in range(16)]
+                 checkHex - In binary, '0bABC', where
+                            A=1: An LVDS input above the IEEE input
+                                 voltage specification limit
+                            B=1: An LVDS input below the IEEE input
+                                 voltage specification limit
```
pomalley: "An LVDS input [was?] above/below..." ?

GHzDACs/Cleanup/dac.py
```diff
@@ -765,21 +782,33 @@ def setPolarity(self, chan, invert):
...
+                            C=1: Sampling in correct data cycle
```
pomalley: What does this mean?

GHzDACs/Cleanup/dac.py
```diff
@@ -765,21 +782,33 @@ def setPolarity(self, chan, invert):
...
+        :rtype: (bool, int, int, int,
+                 tuple(list[int], list[bool], list[bool]),
+                 int)
```
pomalley: It is unfortunate that we have such a convoluted return type...

GHzDACs/Cleanup/dac.py
```diff
@@ -850,6 +879,27 @@ def func():
...
+        the registry entry for this board.
+
+        If no PHOF can be found to get an acceptable FIFO counter, PHOF will
+        be returned as -1, and success will be False.
+
+        :param str chan: Which DAC channel ('A','B')
+        :param int op: 2 for DAC A, 3 for DAC B
+        :param int targetFifo: Desired targetFifo. If None, use value
+                               from board registry entry.
+        :return: found - If procedure successfull
```
pomalley: `sucessful`

GHzDACs/Cleanup/dac.py
```diff
@@ -850,6 +879,27 @@ def func():
...
+                 clkinv - Clock polarity
+                 PHOF - PHOF (phase offset) required for targetFifo.
+                 tries - Number of tries for FIFO counter to equal
+                         targetFifo.
+                 targetFifo - Number bits between read/write signals
+                              for 8-bit buffer.
+        :rtype: (bool, bool, int, int, int)
+        """
```
pomalley: Generally the same comments for this one as for _checkPHOF.

GHzDACs/ghz_fpga_server.py
```diff
@@ -1933,14 +1933,27 @@ def dac_init(self, c, chan, signed=False):
...
+        :param bool optimizeSD: Whether to follow procedure from data
+                                sheet to determine SD
+        :param int/None data: If int, SD value to be set. If None, SD
+                              set to value in registry. Ignored if
+                              optimizeSD=True.
+        :return: success, MSD, MHD, SD, timing profile, checkHex
+                 checkHex - In binary, '0bABC', where
+                            A=1: An LVDS input above specification
+                            B=1: An LVDS input below specification
+                            C=1: Sampling in correct data cycle
```
pomalley: ditto for here as in setLVDS

pomalley:
Thanks for doing this Jim, never had a clue what was going on here. I had some comments so let me know what you think of them.

Also, in general it looks like the lines don't go out to 80 characters--is this the case?

maffoo:
Question for y'all: `dac_lvds` is a labrad setting, so its doc string gets exposed to the world as labrad documentation. The python conventions for documentation may look strange, especially given that this can be accessed from other languages that do things differently. Not saying we should change what @jwenner has done, but just curious what you all think about how to handle this generally.

pomalley:
I have  no idea. :-)

maffoo:
that was my answer too. glad we're in agreement :)

pomalley:
I suppose we should go with the Google python style guide because that is very readable, and keep it to 80 chars because people read it in the console. Other than that... depends on what documentation generator we use, if any?

And you can provide some sort of annotations for the parameters to the setting, right? But IIRC there's some reason we were moving away from that.

maffoo:
The issue is we're using the same docstring for two purposes: (1) documenting the python method, for people reading the python code, and (2) documenting the labrad setting for people calling it from labrad. Some of the info needed for (1) is not relevant for (2), and exactly what is needed for (2) depends somewhat on which language api one is using to call it. But anyway, I don't want to hold up this PR because this is a larger discussion.

jwenner:
I was trying to follow PEP8 on the line length. In particular,
`
For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters. 
`
Should I instead be using 80 characters as the limit?

GHzDACs/Cleanup/dac.py
jwenner: Point taken on the period. However, for the space, in the pyle contributing file we have a space between the `"""` and the start of the text. What do I do?

GHzDACs/Cleanup/dac.py
jwenner: Uh, er, well...

To quote the data sheet:
```
0, phase measurement—sampling in the previous or following DATA cycle.
1, phase measurement—sampling in the correct DATA cycle.
```
Perhaps referring to one of the DATA clocks? But which one (maybe DATACLK_IN, but not sure).

GHzDACs/Cleanup/dac.py
jwenner: Agreed. If I were doing this again, I'd return (success, t, checkHex) and axe the MSD/MHD/cluster garbage since these don't seem to be useful. These come from when the bringup routine actually plotted where MSD and MHD flipped.

However, changing this would require changing not only the server but also the bringup routine & possibly ScanPhases code. I think this is out of scope for this PR. But certainly this should be done at some point for greater clarity in the return (and to clean up that wonderful dac_bringup return tag).

GHzDACs/Cleanup/dac.py
jwenner: Of course, it helps if I'm consistent (cf setLVDS vs setFIFO). Suggestions?

pomalley:
I think limiting comments to 72 characters is silly. PEP8 doesn't really explain why so I'm inclined to stick with 80.

GHzDACs/Cleanup/dac.py
pomalley: you probably saw, but contributing.md got updated w.r.t. this.

jwenner:
@pomalley, could you please take a look now? Tried to implement the changes you suggested. Also added additional explanation on the outputs of the setFIFO + dac_fifo functions. Also converted docstrings in ghz_fpga_server to (I believe) Google format. Thanks for the review!

Since they're hidden as outdated diffs, note my comments above about the LVDS data cycle and return type.

pomalley:
LGTM

PR #202, opened 2015-07-15
Remove un-necessary check on ADC trigger table.

### LINKED ISSUE #201: Trigger table exception in adc.py when running ORBIT scans
I'm raising this issue here because I run into it intermittently when running ORBIT scans and it is (usually) fixed by reducing `m`, leading me to suspect that the problem is in pyle rather than servers. Does anyone know what can cause the following error?

```
Caused by: org.labrad.errors.LabradException: [GHz FPGAs] Remote Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\twisted\internet\defer.py", line 1099, in _inlineCallbacks
    result = g.send(result)
  File "ghz_fpga_server.py", line 607, in run
    pkts = self.makePackets(runners, page, reps, timingOrder, sync)
  File "ghz_fpga_server.py", line 513, in makePackets
    p = runner.setupPacket()
  File "C:\labrad-servers\servers\GHzDACs\Cleanup\adc.py", line 641, in setupPacket
    p = self.dev.setup(self.info)
  File "C:\labrad-servers\servers\GHzDACs\Cleanup\adc.py", line 922, in setup
    self.makeTriggerTable(triggerTable, p)
  File "C:\labrad-servers\servers\GHzDACs\Cleanup\adc.py", line 713, in makeTriggerTable
    raise Exception("Trigger table row0 rcount > 1, this is dangerous as repeating this row accounts for the first rdelay")
exceptions.Exception: Trigger table row0 rcount > 1, this is dangerous as repeating this row accounts for the first rdelay
```

Sometimes I can run the exact same scan and not get it; this makes me think that the problem is running out of SRAM, but... ?

DanielSank:
The check made assumptions about how the trigger table would be written
by higher level code. Those assumptions are not guaranteed, and the check
disallowed valid uses.

Fixes #201


DanielSank:
@pomalley @JulianSKelly

JulianSKelly:
LGTM

DanielSank:
I messed up the commit in a way I don't know how to fix so I'm going to remake this PR.

PR #203, opened 2015-07-15
Remove un-necessesary check in ADC trigger table creation.

DanielSank:
The check made assumptions about how higher level code would write the
trigger table. Since those assumptions are not necessarily correct and
because the check prevented valid use cases, we removed it.

@JulianSKelly for the sake of procedure can you LGTM this again?


JulianSKelly:
LGTM

PR #205, opened 2015-07-17
Usb device fixes

jayich:
We implemented a couple of changes (described in the commits) to get our only two 'USB' GPIB devices working.  

We have found that when using the latest pyvisa that '::INSTR' is included.

```python
>>> import visa
>>> visa.__version__
'1.7'
>>> rm = visa.ResourceManager()
>>> rm.list_resources()
....
```

We also have had problems with devices needing some time between write and read.  This solution looks like a hack.  If you guys have a better solution that would be great.

Credit to @aransfor and @gregllong for both of these fixes.


gpib_server.py
```diff
@@ -107,15 +108,10 @@ def refreshDevices(self):
...
                 try:
-                    if addr.startswith('GPIB'):
-                        instName = addr
-                    elif addr.startswith('TCPIP'):
-                        instName = addr
-                    elif addr.startswith('USB'):
-                        instName = addr + '::INSTR'
-                    else:
+                    device_prefixes = ('GPIB', 'USB', 'TCPIP')
+                    if not (addr.startswith(device_prefixes)):
```
maffoo: didn't know you could pass a tuple to `startswith`. nice.

gpib_server.py
```diff
@@ -145,6 +141,23 @@ def getDevice(self, c):
         instr = self.devices[c['addr']]
         return instr
 
+    def asynchronous_wait(self, seconds, result=None):
```
maffoo: we have a utility function to do exactly this: `labrad.util.wakeupCall`

gpib_server.py
```diff
@@ -176,10 +189,26 @@ def read(self, c, n_bytes=None):
...
         instr = self.getDevice(c)
-        if n_bytes is None:
-            ans = instr.read_raw()
+        if instr.resource_name.startswith('USB'):
+            try:
+                if n_bytes is None:
+                    ans = instr.read()
+                else:
+                    ans = instr.read(n_bytes)
+            except:
```
maffoo: Seems dangerous to have a bare except here. Generally, I think it's preferable to do `except Exception:` so as not to catch things like `KeyboardInterrupt`

gpib_server.py
```diff
@@ -176,10 +189,26 @@ def read(self, c, n_bytes=None):
...
+                self.asynchronous_wait(0.1)
```
maffoo: this would need a `yield` to actually wait for the deferred to fire before continuing:
```python
yield self.asynchronous_wait(0.1)
```
or
```python
yield labrad.util.wakeupCall(0.1)
```

gpib_server.py
```diff
@@ -176,10 +189,26 @@ def read(self, c, n_bytes=None):
...
+                if n_bytes is None:
+                    ans = instr.read()
+                else:
+                    ans = instr.read(n_bytes)
+            ans = str(ans)
+            if ",," in ans:
+                ans = ans.replace(",,", ",")  # rigol specific stupidity
```
maffoo: what's going on here?

gpib_server.py
```diff
@@ -176,10 +189,26 @@ def read(self, c, n_bytes=None):
...
+                if n_bytes is None:
+                    ans = instr.read()
+                else:
+                    ans = instr.read(n_bytes)
+            except:
+                self.asynchronous_wait(0.1)
+                if n_bytes is None:
+                    ans = instr.read()
+                else:
+                    ans = instr.read(n_bytes)
```
maffoo: I'd rather not have this delay/retry logic here in the gpib server, since it seems like a device-specific requirement. Ideally, this would go in the device server for the devices that need it. One thing we might want to do is add a setting on the gpib server that just does a delay, so that then the device server could send a single packet that includes the required delays, e.g.:
```python
p = cxn.gpib_bus.packet()
p.write(msg).delay(0.1).read()
result = yield p.send()
result.read
```
You could also do multiple writes with interleaved delays, etc, but all in one packet to the gpib bus server.

Similarly, we could add a delay parameter to the `query` setting, which would add a delay between the read and write in the same way.

gpib_server.py
jayich: @maffoo , yep, that works.  Superior.  See: https://github.com/martinisgroup/servers/commit/094710dc45d5c8a8bea7cc5addb0c9ac33f32694

aransfor:
I think this is a hack, it's kind of annoying that a specific device server  should have to have some write/read delay when in principle should have some standardized protocol that the GPIB server deals with. I would chalk this up to one poor device but now there is both a Rigol signal generator and some Agilent thing @jayich uses.

gpib_server.py
jayich: @maffoo, okay it appears that everything works without the yield.  We tried with the yield and that ran into errors. The "read" method would likely need to use returnValue instead of return.  I am not sure what else.

aransfor:
I think the GPIB server should at least timeout instead of hanging on a bad read operation

maffoo:
@aransfor, the server is supposed to time out. By default the read timeout is just 1 second, but you can set this in a particular context, if you want to do something for a particular device. I think the `read` and `read_raw` methods might behave differently in this regard, however, so that could be the issue, because I see that the new USB code uses `read`.

gpib_server.py
jayich: I added a clarifying comment in https://github.com/CampbellGroup/servers/commit/66b48359c2ac336a5ea292f17bec261f374338cc.

With this device connected we get the following with pyvisa 1.7.
```python
>>> import visa
>>> rm = visa.ResourceManager()
>>> rigol = rm.get_instrument('USB0::0x09C4::0x0400::DG1D150900538::INSTR')
>>> rigol.write('*IDN?')
>>> rigol.read()
u'RIGOL TECHNOLOGIES,DG1022 ,DG1D150900538,,00.03.00.09.00.02.11\n'
```

The double commas, before 00.03 are the use case we are taking care of.

gpib_server.py
jayich: Included Exception here: https://github.com/CampbellGroup/servers/commit/f58ed8e46092cfc9a768cd289c855817eef73550

PR #206, opened 2015-07-18
Clean up DC rack server

maffoo:
Was looking at this server after talking with @christopherquintana about an issue he was was having, and noticed that it could use some TLC. Cleaned up whitespace, extracted a bunch of inline dicts to top-level constants, and replaced magic numbers in bit manipulations with constants as well.


DCRack/dc_rack_server.py
```diff
@@ -300,42 +367,38 @@ def pushRegistryValue(self, dac, slow, voltage):
...
-        yield self.write([128|(Byte3&63)])
-        yield self.write([128|(Byte2&63)])
-        yield self.write([128|(Byte1&63)])
-        
+            dac_value = long(float(num)/2.5 * 0xffff)
+            reg = (dac_value << 1)
+
+        # set high bit for slew rate
+        if slow:
+            reg |= 0x20000
```
maffoo: There was a nasty bug here before. The expression `intNum | 131702` was meant to set the 18th bit for the slew rate, but `131702 == 0x20276`. It should have been `131072 == 0x20000`. So we were previously messing up the requested DAC levels by turning on a bunch of bits that weren't requested. The moral of the story is never use decimal constants when doing bit manipulations; always use hex or, if the language allows, binary.

DCRack/dc_rack_server.py
```diff
@@ -220,79 +295,71 @@ def identSelf(self, timeout=1):
...
+        ID = CHANNEL_IDS[channel]
+        yield self.write([OP_TRIGGER | ID])
 
 
     @inlineCallbacks
-    def pushRegistryValue(self, dac, slow, voltage):
-        """Pushes 18 bits of data into 18 bit shift register. First bit is fine(0) or coarse(1) DAC, 
-        last bit is fast(0) or slow(1) slew rate, and middle 16 bits are voltage value"""
-        #Conversion of voltage value into 16-bit number, plus bits for DAC selection and slew rate
+    def pushRegisterValue(self, dac, slow, voltage):
```
maffoo: I was confused at first because the code uses 'registry' to refer to the labrad registry in other places, but here we're talking about the data *register* on the DAC card. I've tried to fix up the naming.

DCRack/dc_rack_server.py
christopherquintana: This is great; that bug was causing a serious issue. This branch seems to
be working fine on Jules, so LGTM.

On Fri, Jul 17, 2015 at 6:10 PM, Matthew Neeley <notifications@github.com>
wrote:

> In DCRack/dc_rack_server.py
> <https://github.com/martinisgroup/servers/pull/206#discussion_r34942322>:
>
> > -        #Push bits to proper positions
> > -        Byte1 = long(intNum)
> > -        Byte2 = long(intNum>>6)
> > -        Byte3 = long(intNum>>12)
> > -
> > -        #Write 8 bit sequences
> > -        yield self.write([128|(Byte3&63)])
> > -        yield self.write([128|(Byte2&63)])
> > -        yield self.write([128|(Byte1&63)])
> > -
> > +            dac_value = long(float(num)/2.5 * 0xffff)
> > +            reg = (dac_value << 1)
> > +
> > +        # set high bit for slew rate
> > +        if slow:
> > +            reg |= 0x20000
>
> There was a nasty bug here before. The expression intNum | 131702 was
> meant to set the 18th bit for the slew rate, but 131702 == 0x20276. It
> should have been 131072 == 0x20000. So we were previously messing up the
> requested DAC levels by turning on a bunch of bits that weren't requested.
> The moral of the story is never use decimal constants when doing bit
> manipulations; always use hex or, if the language allows, binary.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/martinisgroup/servers/pull/206/files#r34942322>.
>



-- 
Chris Quintana
Martinis Lab
UC Santa Barbara

PR #212, opened 2015-08-20
Convert top-level qubit server code to scala

maffoo:
See #164

This is the last part of the transliteration phase, with all the code now converted to scala (except for a few enumerations). After this, we'll be able to move on to removing the java-isms to take advantage of the many nice features scala has to offer.


jwenner:
So it doesn't matter that the files in enums haven't been converted?

maffoo:
@jwenner, not sure exactly what you mean. It certainly doesn't matter in the sense that scala can use java enum classes, so everything will continue to work. In the future we may convert these to scala enums, which can be a bit more convenient to use from scala, and can be defined more compactly, but there's no immediate need.

maffoo:
Tested running sequences with real hardware on Jules.

pomalley:
I know I've been rather delinquent with this one, but I've been in 100% all-data-all-the-time mode for a while now. I hope to get to this soon.

QubitServer/src/main/java/org/labrad/qubits/Constants.scala
```diff
@@ -0,0 +1,81 @@
...
+  /*
+   * Maximum number of reps that can be run in one go.
+   *
+   * This number is determined by two factors, first the number
+   * of reps must fit into a two byte field in the packet that
+   * is sent to the DACs, and second it must be a multiple of 30,
+   * since timing results are sent back in groups of 30.
+   */
+  val MAX_REPS = 65535 // No longer has to be a multiple of 30 ERJ
+  //public static final long MAX_REPS = 300;
```
DanielSank: Delete.

QubitServer/src/main/java/org/labrad/qubits/Constants.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  /*
+   * Maximum number of reps that can be run in one go.
+   *
+   * This number is determined by two factors, first the number
+   * of reps must fit into a two byte field in the packet that
+   * is sent to the DACs, and second it must be a multiple of 30,
+   * since timing results are sent back in groups of 30.
+   */
+  val MAX_REPS = 65535 // No longer has to be a multiple of 30 ERJ
```
DanielSank: Delete comment.

QubitServer/src/main/java/org/labrad/qubits/Constants.scala
```diff
@@ -0,0 +1,81 @@
...
+    "SRAM_BLOCK0_LEN" -> 8192,
+    "SRAM_BLOCK1_LEN" -> 2048,
+    "SRAM_WRITE_PKT_LEN" -> 256
+  )
+
+  private def processProperties(props: Map[String, Int]): Data = {
+    val elems = props.map { case (k, v) =>
+      Data.clusterOf(Data.valueOf(k), Data.valueOf(v.toLong))
+    }
+    Data.listOf(elems.toSeq.asJava)
```
DanielSank: What's with the `asJava`?

QubitServer/src/main/java/org/labrad/qubits/Constants.scala
maffoo: done.

QubitServer/src/main/java/org/labrad/qubits/Constants.scala
maffoo: done.

QubitServer/src/main/java/org/labrad/qubits/Device.scala
```diff
@@ -0,0 +1,96 @@
...
+ * @author maffoo
+ */
+class Device(name: String) {
+
+  private val channels: List[Channel] = Lists.newArrayList()
+  private val channelsByName: Map[String, Channel] = Maps.newHashMap()
+
+  def getName(): String = {
+    name
+  }
```
DanielSank: How does this work in Scala? The method returns `name` but we haven't declared that member variable in the class.

QubitServer/src/main/java/org/labrad/qubits/Constants.scala
maffoo: It converts a scala collection type (in this case `scala.collection.Seq`) to a java collection type (`java.util.List`). We have this impedance mismatch because it's much simpler to work with scala collections, but the JLabrad api methods want java collections; the .asJava and .asScala methods do the needed conversions. Once we switch everything over to scalabrad, which can deal with scala collections, these will go away.

QubitServer/src/main/java/org/labrad/qubits/Device.scala
maffoo: Notice the class is defined as `class Device(name: String)`. That defines the constructor for the class, which takes a string, and also makes `name` accessible as a member of the class. It removes several lines of java boilerplate where you have to separately define the field and assign the constructor parameter to it.

DanielSank: Truly, the Scala gods are enlightened and benevolent. This just looks so, so weird compared with other languages.

maffoo: Note that in the next change I have queued up, these `getName` methods just go away entirely, because if you have a public member in scala, you get a generated 'accessor' method with the same name. Then you can just say `device.name` instead of `device.getName()`. I think some of the weirdness here might just be due to the fact that in trying to make these changes incremental (for easier reviewing, I hope), we end up straddling the java/scala worlds in ways that are a little foreign to both languages (see also the collections conversions).

maffoo:
From @DanielSank via chat: "mergy mergy". Going to merge.

DanielSank:
It seems procedure is impeding this project more than it's helping it. I think given that @maffoo already tested the code on Jules and given that we have no tests for this part of the stack, he should merge sans LGTM.

We do *not* encourage this at all. The fact that we're doing this here comes from Scala illiteracy in the group and the fact that @pomalley is trying to graduate. I did look over the code and from what I can tell it's mostly transliteration to scala, so I think it's ok.

Do *not* use this incident as an excuse to skip proper review in the future.

PR #213, opened 2015-09-01
Replace calls to numpy.loadtxt to avoid memory leak

### LINKED ISSUE #198: Data Vault memory leak
screenshot from 2015-06-30 16 55 50

maffoo:
Fixes #198. See notes on that issue about the problems with `numpy.loadtxt`


ejeffrey:
The scalabrad update seems to have fixed the memory leak, so I am closing this.  If the numpy leak exists, it is small enough or doesn't bother us in our current usage.

PR #216, opened 2015-09-10
Fix eval of old parameters saved with 1.#IND for nan.

maffoo:
Depends on labrad/pylabrad#160 to fix the handling of `None` for units. See that PR for more context. This hack is specific to data vault parameters, not needed for backwards compatibility of `evalLRData` itself, which is why I've put it here.


DanielSank:
Yikes, where did that escape sequence come from? Bizarre.

LGTM (although it would be even better with a test).

ejeffrey:
Do we still need this?  If so we should merge it.

PR #218, opened 2015-09-14
fix timeout handling for gpib server and USB device name handling

BrooksCampbell:
GPIB server was not setting the specified timeout setting when doing a read.  Also, for USB devices, it was previously adding a duplicate "::INSTR" at the end of the device address.


gpib_server.py
```diff
@@ -177,8 +177,10 @@ def read(self, c, n_bytes=None):
         """
         instr = self.getDevice(c)
         if n_bytes is None:
+            instr.timeout = c['timeout']['ms']
```
DanielSank: No reason to have this inside the `if` statement, since it's reproduced in the `else`. Just refactor to a single line before the `if/else`.

gpib_server.py
```diff
@@ -177,8 +177,10 @@ def read(self, c, n_bytes=None):
...
             ans = instr.read_raw()
         else:
+            instr.timeout = c['timeout']['ms']
```
DanielSank: See above.

DanielSank:
The timeout change is great. Thanks for doing this @BrooksCampbell!

What's up with removing that bit from the device name string?

BrooksCampbell:
I made your refactoring change.

The device string change isn't very well tested, but both the USB Keysight scopes I've used return their address strings already ending in "::INSTR" so when it was appended to the address a second time I wasn't able to communicate with the instruments (DSOX2014A and DSO90804A are the two scopes I tested this with)...  I'm guessing maybe there are other USB devices that did need this line so maybe it shouldn't be added?  What else uses this part of the server?

pomalley:
A while ago we upgraded pyvisa and it stuck ::INSTR on the end of lots of things. Mostly we just let this be and changed the few servers that actually cared about the device name being returned from *IDN. I guess at some point we someone must have seen that it didn't do this to USB devices and added this to make things consistent, but I don't know any specifics. (This is going from memory, I could be wrong.)

maffoo:
LGTM

DanielSank:
@maffoo are you sure this won't break other stuff?

maffoo:
no, I'm not sure, but it _looks_ good to me...

Seriously, though, I'm happy to defer to @BrooksCampbell here who is actually using gpib over USB with real devices.

jwenner:
@BrooksCampbell, it looks like your latest commit is unrelated to the PR. Is this intentional?

jwenner:
I meant @btchiaro.

BrooksCampbell:
I let this go stale, but would like to merge it.  What do I do now?

maffoo:
@BrooksCampbell, depends on exactly you want to merge. Do you need the WIP changes, e.g. the scope server stuff? If so you should probably get those reviewed separately since they are unrelated to the gpib server changes.

PR #219, opened 2015-09-22
Explain how to install PyQt into a virtualenv on Linux

DanielSank:
PyQt is needed by e.g. the fridge monitor. As it is not packaed for `pip`, we need a manual procedure for installation into a virtualenv. This commit provides that procedure.

It's possible that we could fix this up to work with pip, but I have not investigated that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-servers
-=======
+# servers
+
 A collection of LabRAD servers.  Many of these are for specific pieces of hardware.  In many only 
 have the most basic functionality, or whatever the person who wrote them needed for their experiment.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,3 +11,43 @@ serial_server               Serial port interface using pyserial
...
+  1. `$ cd ~/src/PyQt-x11-gpl-4.11.4`
+  1. `$ python configure-ng.py -q /usr/bin/qmake-qt4`
+  1. `$ make` # This step might take a while.
+  1. `$ sudo make install`
+
+1. Compile and install PyQwt
+  1. `$ cd ~/src/PyQwt-5.2.0/configure`
+  1. `$ python configure.py -Q ../qwt-5.2`
+  1. `$ make -j4`
+  1. `$ sudo make install`
```
maffoo: If I just want SIP, PyQt and PyQwt into my virtual env, do I really need sudo?

README.md
```diff
@@ -11,3 +11,43 @@ serial_server               Serial port interface using pyserial
...
+1. Compile and install SIP
+  1. `$ cd ~/src/sip-4.16.9`
+  1. `$ python configure.py`
+  1. `$ make`
+  1. `$ sudo make install`
+  1. You should see that the relevatn files are copied into your virtualenv.
+
+1. Compile and install PyQt
+  1. `$ cd ~/src/PyQt-x11-gpl-4.11.4`
+  1. `$ python configure-ng.py -q /usr/bin/qmake-qt4`
```
maffoo: why configure-ng.py here instead of just configure.py?

ejeffrey:
Anything that uses setup.py can be made into a binary whl package and installed by pip, but it looks like pyqt does its own thing, so that probably won't work.  It looks like the options are use the ubuntu pre-built versions (which kind of leaves us at their mercy for versions), build our own .deb packages, or build from source on every computer.

README.md
```diff
@@ -11,3 +11,43 @@ serial_server               Serial port interface using pyserial
...
+  1. `$ make` # This step might take a while.
+  1. `$ sudo make install`
+
```
ejeffrey: Why do you use -j4 for PyQwt and not for PyQt?  -j4 is parallel make, it should run a lot faster.

README.md
DanielSank: Yes. Did you try without?

README.md
DanielSank: `configure-ng` is preferred for reasons explained in the PyQt website. I didn't try `configure-ng` for SIP, but it might work. It's not a big deal either way.

README.md
DanielSank: > Why do you use -j4 for PyQwt and not for PyQt?

'cuz I'm dumb. Fixed.

ejeffrey:
LGTM

PR #224, opened 2015-09-29
Remove reference to windows-only getch

maffoo:
Remove imports from msvcrt. The input code was already using `raw_input` anyway (which requires the user to press enter, so this does not change the behavior, just makes it more cross-platform.


DanielSank:
LGTM

PR #226, opened 2015-10-03
Refactor fpga support code into fpgalib package.

maffoo:
It's time for the `Cleanup` package to get cleaned up :) This makes the package structure more like what we have for the datavault server, with the server script in the top-level directory and a package directory (in this case `fpgalib` inside). We assume that the cloned repo directory is placed on the python path, rather than the parent directory, which avoids needing to clone this repo into a subdirectory. There will need to be a corresponding change in pyle to fix imports in fpgaTest.


ejeffrey:
I can't see the failed check from here -- I assume that is some issue with the test script looking in the wrong location, but otherwise LGTM

jwenner:
The failed test info is:
```
fpgalib.test (1)--->test_fpga_server.top_level_collect:
('fpgalib/test/test_fpga_server.py', None, 'fpgalib/test/test_fpga_server.py')
servers/fpgalib/test/test_fpga_server.py:24: in <module>
    import fpgalib.ghz_fpga_server as fpga
E   ImportError: No module named ghz_fpga_server
```

ejeffrey:
Thanks jim.  That should be 'import ghz_fpga_server as fpga'

jwenner:
@maffoo, since you are moving ghz_fpga_server, could you please check (or have me check) after rebasing this on master that #227 is still implemented? Thanks.

jwenner:
Note that you moved the DAC bringup script (GHz_DAC_bringup) into the main servers directory. At least to me, this doesn't make sense. Perhaps it should be moved back to pyle/scripts or to fpgalib/scripts?

jwenner:
Other than these few complaints, LGTM as tested in martinisgroup/pyle#755.
Maybe also bump version number to v5.0.4 (5.0.3 from #227) as other functions access files which have been moved.

maffoo:
Good idea to bump the version number. As for the script location, I think we want to keep the dac bringup script in this repo, but what if I put it into a new `scripts/` directory, similar to the structure we have in the pyle repo?

jwenner:
Having such a scripts directory seems reasonable to me. Your choice on whether to call it scripts or fpgascripts. Plus, if the calibration server/script is cleaned up, this would then be a natural place for the script part.

jwenner:
Note that changing the import signatures here (i.e., removing `servers.`) will affect one more file not included here: ghzdac_cal/GHz_DAC_calibrate.py.

maffoo:
@jwenner, this change doesn't actually impact the calibration script unless on a particular machine you decide to _only_ put the server checkout directory (as opposed to its parent directory) on the python path. I hadn't meant for this change to be an all-encompassing restructuring of the servers repo (along the lines of #109), but rather a step in that direction. However, you bring up a good point that it's probably better to just go all in and change the convention, rather than leaving things in an intermediate state where we have to put multiple things on the python path. Because different server/scripts expect different things.

maffoo:
@DanielSank, I'd like to move ahead on this. As I said above, I'm not attempting a full reorg of the repo as proposed in #109, just introducing one new package `fpgalib` instead of `servers.GHzDACs.Cleanup`. Does this seem ok to you as is?

DanielSank:
Per gchat discussion, LGTM.

PR #237, opened 2015-10-07
ghz_fpgas.findDevices: Put boardGroup removal in try-except

### LINKED ISSUE #235: Board groups not properly removed in ghz_fpga_server when DE server disconnects
From @jwenner in #234:

If I start DE, then start fpga, then close DE, then fpga still thinks the boards are connected. The messages from fpga upon closing DE are:
```
2015-10-06 16:27:30-0700 [ServerProtocol,client] Refreshing client connection...

2015-10-06 16:27:30-0700 [ServerProtocol,client] Loading board group definitions
 from registry...
2015-10-06 16:27:30-0700 [ServerProtocol,client] Board group definitions loaded.

2015-10-06 16:27:30-0700 [ServerProtocol,client] Board group definitions ok.
2015-10-06 16:27:30-0700 [ServerProtocol,client] Adapter "Vince Direct Ethernet"
 (port 3) does not exist. Group will not be added.
2015-10-06 16:27:30-0700 [ServerProtocol,client] Adapter "hercules_direct_ethern
et" (port 2) does not exist. Group will not be added.
2015-10-06 16:27:30-0700 [ServerProtocol,client] Adapter "santo_direct_ethernet"
 (port 0) does not exist. Group will not be added.
2015-10-06 16:27:30-0700 [ServerProtocol,client] Adapter "DR Direct Ethernet" (p
ort 0) does not exist. Group will not be added.
2015-10-06 16:27:30-0700 [ServerProtocol,client] Adapter "ivan_direct_ethernet"
(port 0) does not exist. Group will not be added.
2015-10-06 16:27:30-0700 [ServerProtocol,client] Adapter "runningman_direct_ethe
rnet" (port 1) does not exist. Group will not be added.
2015-10-06 16:27:30-0700 [ServerProtocol,client] Adapter "rottmayer_direct_ether
net" (port 2) does not exist. Group will be removed.
2015-10-06 16:27:30-0700 [ServerProtocol,client] Board groups to be added: set([
])
2015-10-06 16:27:30-0700 [ServerProtocol,client] Board groups to be removed: set
([('rottmayer_direct_ethernet', 2L)])
2015-10-06 16:27:31-0700 [ServerProtocol,client] Unhandled error in Deferred:
2015-10-06 16:27:31-0700 [ServerProtocol,client] Unhandled Error
        Traceback (most recent call last):
        Failure: labrad.types.Error: (0) no such server: 5 [payload=None]
```
At this point, `ghz_fpgas.list_devices()` will still list the devices before closing DE. Upon restarting DE, the list of boards will be updated to what is now connected.

jwenner:
When scala DirectEthernet server closed while ghz_fpgas still running, boards on that DirectEthernet server were never removed due to never finding server. Place closing of boardGroup in try-except structure so that, if server not found, will continue on and remove boards on the DirectEthernet server.

Fixes #235


GHzDACs/ghz_fpga_server.py
```diff
@@ -990,7 +991,13 @@ def findDevices(self):
         for key in removals:
             bg = self.boardGroups[key]
             del self.boardGroups[key]
-            yield bg.shutdown()
+            try:
+                yield bg.shutdown()
+            except labrad.types.Error, e:
+                if 'no such server' in str(e):
```
maffoo: This kind of check that relies on the content of an error message is really brittle, and generally not something you want to rely on for correctness. I would suggest instead that we add a check in the shutdown method itself to see whether the server still exists and skip the expire_context call if not, and here just capture any exception and log it.

```python
# in BoardGroup:
@inlineCallbacks
def shutdown(self):
    cxn = self.directEthernetServer._cxn
    servers = yield cxn.manager.servers()
    server_ids = set(id for id, name in servers)
    if self.directEthernetServer.ID in ids:
        yield cxn.manager.expire_context(
                self.directEthernetServer.ID, context=self.ctx)

...
# here
for key in removals:
    bg = self.boardGroups[key]
    del self.boardGroups[key]
    try:
        yield bg.shutdown()
    except Exception, e:
        logging.error('Exception in BoardGroup shutdown: {}'.format(e))
```

One of the things that people have raised before is having a better way to deal with errors based on error codes, which would be much better than matching strings in error messages. That's something we could explore, but clearly outside the scope of this change.

maffoo:
fixed vias #247

PR #240, opened 2015-10-12
Update serial server to work on Linux as well as windows (also OSX)

ejeffrey:
We were using hard-coded COM port names from COM1..COM20.  This worked on Windows, but not on Linux or OSX.  pyserial has a built-in device name enumeration, but it might not work properly on windows.  So we keep the same detection code on windows, but use the pyserial built-in stuff on Linux and OSX.  I attempted to not change the windows code path, but don't have a good way to test it.  Can someone run this on windows and verify that list_serial_ports() and open() work properly?

### DIFF
--- a/serial_server.py
+++ b/serial_server.py
@@ -17,7 +17,7 @@
 ### BEGIN NODE INFO
 [info]
 name = Serial Server
-version = 1.0
+version = 1.1
 description = 
 instancename = %LABRADNODE% Serial Server
 
--- a/serial_server.py
+++ b/serial_server.py
@@ -31,11 +31,11 @@ timeout = 20
 ### END NODE INFO
 """
 
-import sys
-import os
-from time import sleep
-import time
 import collections
+import os
+import sys
+import time
+from time import sleep
 
 from labrad import types as T
 from labrad.errors import Error
@@ -56,7 +56,9 @@ class NoPortsAvailableError(Error):
     """No serial ports are available."""
     code = 3
 
+
 SerialDevice = collections.namedtuple('SerialDevice', ['name', 'devicepath'])
+
 
 class SerialServer(LabradServer):
     """Provides access to a computer's serial (COM) ports."""
@@ -69,7 +71,8 @@ class SerialServer(LabradServer):
             self.enumerate_serial_pyserial()
             
     def enumerate_serial_windows(self):
-        """This manually attempts to enumerate the first 20 COM ports.
+        """Manually Enumerate the first 20 COM ports.
+
         pyserial includes a function to enumerate device names, but it
         possibly doesn't work right on windows for COM ports above 4.
         http://stackoverflow.com/questions/12090503/listing-available-com-ports-with-python
@@ -91,10 +94,12 @@ class SerialServer(LabradServer):
             print '  none'
 
     def enumerate_serial_pyserial(self):
-        """This uses the pyserial built-in device enumeration.  However,
-        we ignore the pyserial "human readable" device name because that
-        appears to make no sense.  For instance, a particular FTDI USB-Serial
-        adapter shows up as 'Microsoft Corp. Optical Mouse 200'.
+        """This uses the pyserial built-in device enumeration.
+
+        We ignore the pyserial "human readable" device name
+        because that appears to make no sense.  For instance, a
+        particular FTDI USB-Serial adapter shows up as 'Microsoft
+        Corp. Optical Mouse 200'.
 
         Following the example from the above windows version, we try to open
         each port and ignore it if we can't.
@@ -136,7 +141,7 @@ class SerialServer(LabradServer):
         return port_list
 
     @setting(10, 'Open',
-             port=[': Open the first available port', # Is this really useful??? - ERJ
+             port=[': Open the first available port',
                    's: Port to open, e.g. COM4'],
              returns=['s: Opened port'])
     def open(self, c, port=''):
@@ -159,9 +164,7 @@ class SerialServer(LabradServer):
                 if x.name.lower() == port.lower():
                     try:
                         c['PortObject'] = Serial(x.devicepath)
-                        # Strip off the leading \\.\ if it exists for
-                        # windows compatiblity
-                        return c['PortObject'].portstr.replace('\\\\.\\', '')
+                        return x.name
                     except SerialException, e:
                         if e.message.find('cannot find') >= 0:
                             raise Error(code=1, msg=e.message)
--- a/serial_server.py
+++ b/serial_server.py
@@ -31,8 +31,11 @@ timeout = 20
 ### END NODE INFO
 """
 
+import sys
+import os
 from time import sleep
 import time
+import collections
 
 from labrad import types as T
 from labrad.errors import Error
@@ -42,7 +45,7 @@ from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.internet.task import deferLater
 from serial import Serial
 from serial.serialutil import SerialException
-
+import serial.tools.list_ports
 
 class NoPortSelectedError(Error):
     """Please open a port first."""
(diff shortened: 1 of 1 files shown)


serial_server.py
```diff
@@ -31,8 +31,11 @@
 ### END NODE INFO
 """
 
+import sys
+import os
 from time import sleep
 import time
+import collections
```
maffoo: imports should be sorted. I would also suggest removing the bare import of `sleep` and just referring to `time.sleep`, though that may be out of scope for this change.

serial_server.py
```diff
@@ -53,28 +56,62 @@ class NoPortsAvailableError(Error):
...
     name = '%LABRADNODE% Serial Server'
 
     def initServer(self):
+        if sys.platform.startswith('win32'):
+            self.enumerate_serial_windows()
+        else:
+            self.enumerate_serial_pyserial()
+            
+    def enumerate_serial_windows(self):
+        """This manually attempts to enumerate the first 20 COM ports.
```
maffoo: doc style: blank line after first line of dic string.

serial_server.py
```diff
@@ -53,28 +56,62 @@ class NoPortsAvailableError(Error):
     """No serial ports are available."""
     code = 3
 
+SerialDevice = collections.namedtuple('SerialDevice', ['name', 'devicepath'])
```
maffoo: style guid says 2 blank lines between top-level definitions.

serial_server.py
```diff
@@ -53,28 +56,62 @@ class NoPortsAvailableError(Error):
...
+        possibly doesn't work right on windows for COM ports above 4.
+        http://stackoverflow.com/questions/12090503/listing-available-com-ports-with-python
+        """
         self.SerialPorts = []
         print 'Searching for COM ports:'
         for a in range(1, 20):
             COMexists = True
             try:
-                ser = Serial('\\\\.\\COM%d' % a)
+                ser = Serial(r'\\.\COM%d' % a)
```
maffoo: I would suggest constructing strings first, then using them throughout:
```python
name = 'COM{}'.format(a)
devicepath = r'\\.\' + name
try:
    ser = Serial(devicepath)
...
```

maffoo: also, might be best to pull `r'\\.\'` out as a global like `WIN_DEVICEPATH_PREFIX`

serial_server.py
```diff
@@ -53,28 +56,62 @@ class NoPortsAvailableError(Error):
...
                     COMexists = False
             if COMexists:
-                self.SerialPorts += ['COM%d' % a]
+                self.SerialPorts.append(SerialDevice('COM%d' % a, r'\\.\COM%d' % a))
                 print '  COM%d' % a
         if not len(self.SerialPorts):
             print '  none'
 
+    def enumerate_serial_pyserial(self):
+        """This uses the pyserial built-in device enumeration.  However,
```
maffoo: Doc style: blank line after first line of doc string.

serial_server.py
```diff
@@ -53,28 +56,62 @@ class NoPortsAvailableError(Error):
...
+        we ignore the pyserial "human readable" device name because that
+        appears to make no sense.  For instance, a particular FTDI USB-Serial
+        adapter shows up as 'Microsoft Corp. Optical Mouse 200'.
+
+        Following the example from the above windows version, we try to open
+        each port and ignore it if we can't.
+        """
+        dev_list = serial.tools.list_ports.comports()
+        self.SerialPorts = []
```
maffoo: would be nice to rename this as `serial_ports`, but that may be out of scope for this change.

serial_server.py
```diff
@@ -93,10 +130,13 @@ def list_serial_ports(self, c):
...
         including ones that are already in use by other programs."""
-        return self.SerialPorts
+        print self.SerialPorts
+        port_list = [x.name for x in self.SerialPorts]
+
+        return port_list
 
     @setting(10, 'Open',
-             port=[': Open the first available port',
+             port=[': Open the first available port', # Is this really useful??? - ERJ
```
maffoo: This is probably not useful, but let's have the discussion here or in a separate issue instead of inline comments. I would be fine to remove.

serial_server.py
```diff
@@ -108,21 +148,27 @@ def open(self, c, port=''):
...
-            try:
-                c['PortObject'] = Serial('\\\\.\\' + port, timeout=0)
-            except SerialException, e:
-                if e.message.find('cannot find') >= 0:
-                    raise Error(code=1, msg=e.message)
-                else:
-                    raise Error(code=2, msg=e.message)
-        return c['PortObject'].portstr.replace('\\\\.\\', '')
+            for x in self.SerialPorts:
+                if x.name.lower() == port.lower():
```
maffoo: Why make this case insensitive? If we want that, it should be noted in the setting documentation that we use case-insensitive comparison for the port name, however I would vote against unless there's a compelling reason to do so.

serial_server.py
```diff
@@ -108,21 +148,27 @@ def open(self, c, port=''):
...
+                    try:
+                        c['PortObject'] = Serial(x.devicepath)
+                        # Strip off the leading \\.\ if it exists for
+                        # windows compatiblity
+                        return c['PortObject'].portstr.replace('\\\\.\\', '')
```
maffoo: Can use raw string literal as above, or, better, a global constant for the prefix to be replaced. Also, might want to limit this replacement to only strip prefixes, instead of replacing the string everywhere.

serial_server.py
ejeffrey: because device names as case insensitive on windows (com1 and COM1 are equally valid).  Previously, it was implicitly case insensitive because it just passed whatever name you passed to open (after prepending \\.\).  Here, we are matching the name against a list, so this keeps the same behavior.

serial_server.py
```diff
@@ -108,21 +148,27 @@ def open(self, c, port=''):
...
+                    except SerialException, e:
+                        if e.message.find('cannot find') >= 0:
```
maffoo: I think we should try to avoid parsing error message strings, because they can easily change. And given that codes 1 and 2 don't really have any defined meanings anywhere, I'm not sure this helps much. We definitely need a more structured error-handling scheme...

serial_server.py
```diff
@@ -108,21 +148,27 @@ def open(self, c, port=''):
...
+                            raise Error(code=1, msg=e.message)
+                        else:
+                            raise Error(code=2, msg=e.message)
+        raise Error(code=1, msg='Unknown port %s' % (port,))
```
maffoo: prefer the `format` method to `%`-formatting

serial_server.py
maffoo: Ok, I see code 1 is used at the top of the file for `NoPortSelectedError`. If we want to make distinctions here, would probably be best to define more error subclasses like that.

serial_server.py
```diff
@@ -54,27 +57,66 @@ class NoPortsAvailableError(Error):
...
 
+SerialDevice = collections.namedtuple('SerialDevice', ['name', 'devicepath'])
+
+
 class SerialServer(LabradServer):
     """Provides access to a computer's serial (COM) ports."""
     name = '%LABRADNODE% Serial Server'
 
     def initServer(self):
+        if sys.platform.startswith('win32'):
```
maffoo: not sure, but AFAICT the canonical way to check this is to use the `platform.system` function (https://docs.python.org/2/library/platform.html#platform.system):
```python
import platform
...
if platform.system() == 'Windows':
```

serial_server.py
maffoo: Makes sense; I doubt we're actually relying on case insensitivity anywhere, but I'm ok to leave it. It would be good to add this to the doc string, however.

ejeffrey: Changing this to use os.path.normcase which acts like 'lower' on windows, and identity on linux.

thchwhite:
Verified that it works in Windows.  LGTM

maffoo:
One other thing: be sure to bump the version number. Otherwise LGTM

PR #242, opened 2015-10-13
Fix javaisms in qubit server code to more idiomatic scala

maffoo:
See #164 

This contains 3 commits to remove various remaining javaisms in the code: using 'name' instead of 'getName' for name accessor, using scala collections and using scala futures instead of java futures.


DanielSank:
Tested a few pulse sequences on Jules setup. Looked ok.

Tried various combinations of pi pulses and readout pulses. For the most part everything looked ok. There was an occurrence of a malformed readout pulse, but I saw the same thing on Vince last night so I doubt this problem is introduced by this PR.

PR #246, opened 2015-10-20
Use LabradException to avoid verbose tracebacks.

### LINKED ISSUE #233: Direct Ethernet server timeout errors contain overly-verbose stack trace
In operations like "read" where timeouts are expected, the stack trace doesn't help, and in fact mostly just scares and confuses users. We should get rid of these stack traces. See #232 for an example.

maffoo:
Fixes #233

### DIFF
--- a/DirectEthernet/build.sbt
+++ b/DirectEthernet/build.sbt
@@ -2,7 +2,7 @@ organization := "org.labrad"
 
 name := "DirectEthernet"
 
-version := "1.2.1"
+version := "1.2.2"
 
 scalaVersion := "2.11.7"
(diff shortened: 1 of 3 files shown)


ejeffrey:
So this looks fine, but I don't understand it. I get catching and throwing a new exception for timeouts to avoid including the whole stack trace on something that isn't exceptional, but why does replacing sys.error() with fail() improve things?

maffoo:
`fail` is just a convenience method to avoid writing `throw new LabradException` everywhere. The scalabrad api does not include a stack trace for `LabradException`, but does so for any other exception type, including `RuntimeException` which is what `sys.error` throws.

ejeffrey:
LGTM

PR #248, opened 2015-10-21
Update datavault to support TLS

ejeffrey:
Tested with both pre- and post-TLS pylabrad.  The multihead datavault needs to be updated because it doesn't use runServer()


maffoo:
LGTM

PR #249, opened 2015-10-22
Fix timeout recovery to work with or without executionCount

maffoo:
@christopherquintana


DanielSank:
Setting the execution count to a string seems awkward. How about `None` instead.

maffoo:
That's `getattr`, not `setattr`.

DanielSank:
Touche. LGTM.

ejeffrey:
Why not initialize execution count to None rather than have to use the getattr with a default value?

PR #250, opened 2015-10-29
Convert qubit server to use scalabrad

maffoo:
This largely completes the migration of the qubit server to scala, including using the scalabrad api. After this, we can begin the real work of actually improving the server architecture, removing cruft, and building a new gate compiler scheme.


pomalley:
Well it was a lot of Scala all at once after I hadn't looked at any for a month or more, but LGTM. Pending tests on a fridge, I suppose.

pomalley:
Also, should move the source files under main/src/scala instead of main/src/java?

maffoo:
Yeah, I'll add a commit to move things as a final step.

maffoo:
Tested this with @btchiaro on the fridge; fixed one issue with channel identifiers in timing order, but otherwise everything worked fine.

DanielSank:
YES!

PR #255, opened 2015-11-07
Print mks errors. Change mks readout time.

jwenner:
Add print statements to mks server for when returning NaN or min/max scale. When invalid data received, would set reading to one of these without indicating cause of error. If error present, also
print the pressure and full_scale readings so user can see what's happening.

With a 0.02s readout time, the mks server would start getting back corrupted data after a few minutes; this would then continue on as the buffer would not be cleared. Change readout time to 0.05s, which seems to be enough to fully read data out without leading to data corruption.

### DIFF
--- a/mks_server.py
+++ b/mks_server.py
@@ -17,7 +17,7 @@
 ### BEGIN NODE INFO
 [info]
 name = MKS Gauge Server
-version = 2.0
+version = 2.0.1
 description = 
 
 [startup]
@@ -32,6 +32,12 @@ timeout = 20
 
 import labrad 
 import math
+import logging
+# The logging level is set at the bottom of the file where the server starts.
+# To get additional info about what the server is doing (i.e. to see any failed
+# readings and what led to failures), change the logging level to logging.DEBUG
+# and run in separate command prompt.
+
 from labrad import types as T, util
 from labrad.server import LabradServer, setting
 
@@ -154,7 +160,7 @@ class MKSServer(LabradServer):
                 G['ready'] = False
 
             # create a packet to read the pressure
-            pause_len = T.Value(0.02, 's')
+            pause_len = T.Value(0.05, 's')
             p = ser.packet(context=ctx)\
                    .write_line('p').pause(pause_len)\
                    .read_line(key='pressure').pause(pause_len)\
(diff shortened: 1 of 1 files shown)


DanielSank:
I vote for logging instead of print statements. Can anyone think of an example for Jim to follow?

ejeffrey:
The node already collects stdout and stderr and directs them to a log, with priority info for stdout and warn for stderr.  I don't think there is a need for individual servers to use log handling.  They already do use twisted's log filter which intercepts stdout and slaps on a timestamp and other information, but I actually think we might want to get rid of that.

ejeffrey:
But maybe change the error messages to go to stderr with:

print >>sys.stderr, 'Serial FS split error:', e

jwenner:
@ejeffrey, I've now switched to using logging.debug so that the errors don't automatically show up each time in the node but can be viewed in command prompt (if logging level changed). Note that this is akin to ghz_fpga_server. So, how does this look now? Thanks.

jwenner:
@DanielSank, @zchen088, could one of you potentially take a look at this? Note that it's possible (but as yet untested) that this could help #271, since the increase in the timeout time (in the second commit) fixed our mks readout.

mks_server.py
```diff
@@ -212,12 +226,19 @@ def update_readings(self):
...
                     else:
                         data2 = float(rdg2)
-                except ValueError:
+                except ValueError as e:
+                    logging.debug('Serial reading 2 error: {}'.format(e))
+                    error = True
                     data2 = float('NaN')
+                if error:
+                    logging.debug('Pressure: {}'.format(result.pressure))
+                    logging.debug('Full Scale: {}'.format(result.full_scale))
```
DanielSank: I'd add a log message saying that there was, in fact, an error. Otherwise it's not entirely obvious why these values are being logged, particularly since they'll be messed up by construction.

DanielSank:
@jwenner This mostly looks good. Can you add a bit to the logging to indicate to the hapless user than an error actually happened?

jwenner:
@DanielSank, is this what you wanted, or did you want it at a higher level (e.g., info or warn) that would always print to the node?

jwenner:
@DanielSank, does this look good now? Thanks!

jwenner:
bump @DanielSank, is this the type and level of message that you wanted?

jwenner:
bump @DanielSank. Anything else this needs?

jwenner:
@ejeffrey or @maffoo, since @DanielSank doesn't seem to be getting back to me, could one of you take a look? Thanks.

mks_server.py
```diff
@@ -212,12 +226,20 @@ def update_readings(self):
...
                         data2 = float(rdg2)
-                except ValueError:
+                except ValueError as e:
+                    logging.debug('Serial reading 2 error: {}'.format(e))
+                    error = True
                     data2 = float('NaN')
+                if error:
+                    logging.debug('Error occured. Values:')
+                    logging.debug('    Pressure: {}'.format(result.pressure))
+                    logging.debug('    Full Scale: {}'.format(result.full_scale))
```
maffoo: It's better to create one multiline string and then make one logging call. If you make separate calls it's much harder to piece the log together because these statements may get interleaved with other messages, especially when using syslog.

mks_server.py
```diff
@@ -184,17 +190,23 @@ def update_readings(self):
         for G in self.gauges:
             try:
                 result = yield G['packet'].send()
-            except Exception:
+            except Exception as e:
+                logging.debug('Serial return error: {}'.format(e))
```
maffoo: to log exceptions, you can just add an `exc_info` keyword arg (see [here](https://docs.python.org/2/library/logging.html#logging.Logger.debug)):
```python
except Exception:
    logging.debug('Serial return error', exc_info=True)
```

jwenner:
@maffoo, how does this look now? Thanks!

maffoo:
LGTM

PR #256, opened 2015-11-07
DACcal: Add sampling scope channel registry key

jwenner:
Make sampling scope channel a registry key instead of hardcode. Done in case scope channels break.


jwenner:
This PR is now ready to be reviewed. It makes the following changes to the DC Pulse DAC calibration:
1. Make sampling scope channel a registry key instead of hardcode. Done in case scope channels break.
2. After DC pulse, a null trace is output to zero output signal. With prior length, FPGA server would error as length < 20ns. Increase length of this signal to be > 20ns.
3. Make sampling scope sensitivity a registry key instead of hardcode. Done as pulse was going off-screen.

jwenner:
@maffoo or @DanielSank, could you please take a look? Charles will be needing this for his board cals. Thanks!

maffoo:
Same comment applies as the last similar PR: when adding new registry keys it's best to have proper defaults on `reg.get` so that this won't break on older registry data. Otherwise LGTM.

jwenner:
@maffoo, does this look like the correct format for the registry keys? If so, can I go ahead and merge? I've tested this on actual hardware (at least except for the latest commit of providing the defaults). Thank you very much!

ghzdac/calibrate.py
```diff
@@ -346,7 +346,7 @@ def calibrateACPulse(cxn, boardname, baselineA, baselineB, use_switch=True):
     scope.select_device(scopeID)
     p = scope.packet().\
     reset().\
-    channel(SCOPECHANNEL).\
+    channel(reg.get(keys.SSCOPECHANNEL, 'w', True, 2)).\
```
maffoo: I would suggest omitting the pattern here because if you set `2` it will get flattened as `i` which will then fail if you try to load it with the pattern `w`. The way labrad treats integral values is not as flexible as python, unfortunately. But you can omit the pattern and just do `reg.get(keys.SSCOPECHANNEL, True, 2)`.

ghzdac/calibrate.py
```diff
@@ -444,11 +444,11 @@ def calibrateDCPulse(cxn,boardname,channel):
     p = scope.packet().\
     select_device(scopeID).\
     reset().\
-    channel(SCOPECHANNEL).\
+    channel(reg.get(keys.SSCOPECHANNEL, 'w', True, 2)).\
```
maffoo: ditto here

ghzdac/calibrate.py
```diff
@@ -444,11 +444,11 @@ def calibrateDCPulse(cxn,boardname,channel):
...
     trace(1).\
     record_length(5120).\
     average(128).\
-    sensitivity(Value(100.0,'mV')).\
+    sensitivity(reg.get(keys.SSCOPESENSITIVITYDC, 'v', True, 200*labrad.units.mV)).\
```
maffoo: can omit the pattern, or else I would suggest including units: `v[mV]`. The value should get converted properly, for example if the registry value is in volts instead of millivolts (though we should test that).

ghzdac/calibrate.py
```diff
@@ -465,7 +465,7 @@ def calibrateDCPulse(cxn,boardname,channel):
     trace = trace[trace.unit]  # strip units
     # set the output to zero so that the fridge does not warm up when the
     # cable is plugged back in
-    fpga.dac_run_sram([makeSample(dac_neutral, dac_neutral)]*4,False)
+    fpga.dac_run_sram([makeSample(dac_neutral, dac_neutral)]*20,False)
```
maffoo: whitespace after comma

maffoo:
A few comments, but otherwise LGTM.

PR #257, opened 2015-11-08
DACCal: Change length of final DC Pulse signal

jwenner:
After DC pulse, a null trace is output to zero output signal. With prior length, would error as length < 20ns. Increase length of this signal to be > 20ns.


jwenner:
Closing to combine with #256 and another change for DC Pulse cal

PR #260, opened 2015-11-10
AC Pulse Cal - Allow for no switch

jwenner:
The AC Pulse calibration for the DACs assumes the use of a microwave switch. This PR adds a registry key to determine if a switch is present.

In addition, the final reset length was <20ns, the minimum jump table sequence length. This length is increased.

### DIFF
--- a/ghzdac_cal/GHz_DAC_calibrate.py
+++ b/ghzdac_cal/GHz_DAC_calibrate.py
@@ -76,25 +76,28 @@ def calibrate_dc_pulse(fpga_name, channel, dc_scope):
             raise Exception("invalid scope: {}".format(dc_scope))
 
 
-def zero_fixed_carrier(fpga_name):
+def zero_fixed_carrier(fpga_name, use_switch=True):
     """ Calls ghzdac.calibrate.zeroFixedCarrier, synchronously.
     :param str fpga_name: e.g. "Vince DAC 11"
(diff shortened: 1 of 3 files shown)


ghzdac_cal/GHz_DAC_calibrate.py
```diff
@@ -120,8 +123,10 @@ def calibrate_pulse(cxn, fpga_name, dc_scope = 'infiniium'):
         board_type = 'dc'
     
     if board_type == 'ac':
-        zero_a, zero_b = zero_fixed_carrier(fpga_name)
-        dataset = calibrate_ac_pulse(fpga_name, zero_a, zero_b)
+        use_switch = reg.get(keys.SWITCHUSE)
```
maffoo: It would be good to set registry key to the default value in case it does not exist:
```python
use_switch = reg.get(keys.SWITCHUSE, 'b', True, True)
```
The third argument `True` means "set if the key does not exist" and the fourth argument is the default value to set, which just happens to be the same as the third arg in this case.

jwenner:
@maffoo, could you please take a look now? I've made your recommended change. Thanks!

maffoo:
LGTM. has this been tested?

jwenner:
I verified that I can now run the AC Pulse cals (which I couldn't do before). So am I good to merge in?

maffoo:
go for it

PR #261, opened 2015-11-12
DACCal.IQ: Handle Pyvisa errors

jwenner:
When the spectrum analayzer resets itself (lots of clickies), it was timing out, causing the IQ calibration to fail. Place call to spectrum analyzer in try/except loop so error can be handled gracefully.

### DIFF
--- a/ghzdac/calibrate.py
+++ b/ghzdac/calibrate.py
@@ -20,6 +20,7 @@
 
 import time
 import numpy as np
+import pyvisa
 import labrad
 from labrad.types import Value
 import keys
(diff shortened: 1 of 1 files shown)


pomalley:
I know it's been a while on this but if you're interested, Jim, we can get this in. Have you tested this? (I'm guessing yes.) I might suggest going for a limited number of tries (5 or 10?) rather than going forever--if something is wrong with the VISA connection rather than just taking too long we don't want to be stuck in a loop.

ejeffrey:
Yes. I agree, there should be a hard timeout in the loop.

FWIW, there are some alternatives to this, but I think they may be more trouble than they are worth.  You can disable auto align while a scan is running, but then you have to poll the device and run the alignment yourself.  Alternately, you can query if the device is performing an auto-alignment and avoid the timeout that way, although you have to be careful about race conditions.

According to the manual, one of the triggering causes for the auto align cycle is temperature drift of the instrument, so keeping the door closed on the lab may help.

PR #263, opened 2015-11-13
Document Get Data Raw in the qubit sequencer

### LINKED ISSUE #262: Document return signature of qubit sequencer's .run method
Talking to Jimmy just now, I learned that only a few people in the lab understand the return signature of this incredibly important method. Surprised, I looked at the LabRAD doc and found it's almost completely blank. This is likely responsible for a feeling of mystery surrounding how to process data returned by the sequencer. In Jimmy's own words: "I had no idea about the channels, states, demod thing until ..."

Not good :P

DanielSank:
Fixes #262.

@maffoo I think you're the only person around here qualified to judge scala code :P

### DIFF
--- a/QubitServer/src/main/java/org/labrad/qubits/QubitContext.scala
+++ b/QubitServer/src/main/java/org/labrad/qubits/QubitContext.scala
@@ -1117,7 +1117,17 @@ class QubitContext extends AbstractServerContext {
 
   @Setting(id = 1100,
       name = "Get Data Raw",
-      doc = "Gets the raw timing data from the previous run. Given a deinterlace argument, only DAC data are returned")
+      doc = """Get the raw timing data from the previous run.
+              |
+              |The returned data is a 4 index array. The meanings of the indices
+              |are (channel, stat, demod, IQ):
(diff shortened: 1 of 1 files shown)


DanielSank:
By the way, I used all the `|` characters because I saw that done elsewhere in the code, but it looks dorky in the python interactive shell. What's the reason for those?

maffoo:
LGTM. The pipe characters are there to mark how much of the line is just whitespace for formatting the string literal in the source file, versus what should be in the string literal itself. A method called `unwrap` strips these out. Scalabrad calls this method, jlabrad does not. So once we finish the scalabrad conversion, these will magically disappear when looking at the docs in ipython or wherever else.

DanielSank:
@maffoo is there a way to associate docstrings to scala methods/functions and then use them for the LabRAD string, similarly to what we do in pylabrad?

maffoo:
@DanielSank, there's no way to reflectively access scaladoc strings at runtime, that's why you have to put the docs in annotations instead to get them to show up in labrad.

DanielSank:
@maffoo That's too bad. Perhaps we should use a simple docstring registry with it's own decorator?

PR #279, opened 2015-12-03
Fix timeout recovery code in ghz_fpga_server.py

maffoo:
The timeout recovery was failing due to the run context getting prematurely unlocked during timeout recovery. This was causing register pings to be interleaved with the next sequence run, causing cascading timeout failures that could bring scans to a halt. We fix this by waiting until after all boards have been pinged to send any triggers for failed boards.

In addition to the logic fix, also includes some minor cleanups of the timeout recovery and reporting code.

Dupe of #278 since that got merged accidentally before review

### DIFF
--- a/ghz_fpga_server.py
+++ b/ghz_fpga_server.py
@@ -177,7 +177,7 @@ cmdTime_cycles does not properly estimate sram length
 ### BEGIN NODE INFO
 [info]
 name = GHz FPGAs
-version = 5.1.0
+version = 5.2.0
 description = Talks to DAC and ADC boards
 
 [startup]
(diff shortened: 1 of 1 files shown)


ghz_fpga_server.py
```diff
@@ -835,61 +835,44 @@ def extractTiming(self, packets):
...
-        #       dac.py
         for runner, (success, result) in zip(runners, results):
-            ctx = None if success else self.ctx
-            # 1. Clear packet buffer for this board
             yield runner.dev.clear().send()
-            # 2. Ping the board to read its execution counter
             try:
+                # NOTE: in the current implementation of regPing for DAC boards
+                # (build 15) the start field is set to master, which means when
+                # we ping these boards they will emit daisy chain signals.
```
DanielSank: "they will emit" -> "they emit"?

ghz_fpga_server.py
```diff
@@ -835,61 +835,44 @@ def extractTiming(self, packets):
...
                 p = runner.dev.regPingPacket()
-                p.timeout(U.Value(1.0, 's')).collect(1).read(1)
+                p.timeout(U.Value(1.0, 's')).read(1)
```
DanielSank: I'd chain this to the previous line, or write three separate lines, i.e. keep it consistent.

ghz_fpga_server.py
```diff
@@ -835,61 +835,44 @@ def extractTiming(self, packets):
...
-            # failed board
-            finally:
-                yield runner.dev.clear().send()
-                if ctx is not None:
-                    yield runner.dev.trigger(ctx).send()
+                regs = runner.dev.processReadback(resp.read[0][3])
+                runner.executionCount = regs.get('executionCounter', None)
+            except Exception:
+                logging.error('Exception in recoverFromTimeout', exc_info=True)
+        for runner, (success, result) in zip(runners, results):
```
DanielSank: Note that this is the second step?

DanielSank:
A few small things but LGTM.

PR #280, opened 2015-12-03
Add wireshark plugin for analyzing FPGA network traffic.

maffoo:
This proved useful in tracking down the bug fixed in #279. There's still a lot that could be added--for example I haven't defined all the protocol fields and it would be nice to add some summary packet info in addition to the dissection tree--but I thought it'd be nice to commit this and others can contribute.

### DIFF
--- a/fpgalib/tools/ghz_fpgas.lua
+++ b/fpgalib/tools/ghz_fpgas.lua
@@ -1,0 +1,259 @@
+-- Protocol definitions and packet dissectors for GHz FPGAs
+--
+-- To use this in wireshark, you can copy this file into your local wireshark
+-- plugin directory. You can find the plugin directory for your platform by
+-- looking at the wireshark help menu. This is typically something like:
+--
+--  windows: C:/Users/<user>/AppData/Roaming/Wireshark/plugins/
+--  linux: ~/.wireshark/plugins/
+--
+-- Note that this uses heuristic dissectors to identify the fpga packets.
+-- Heuristic dissectors were first exposed to the lua API in version 1.11.3.
+-- On some linux distributions, the packaged version of wireshark is too old,
+-- so you may have to build from source.
+--
+-- The basic structure here is as follows:
+--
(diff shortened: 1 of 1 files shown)


fpgalib/tools/ghz_fpgas.lua
```diff
@@ -0,0 +1,206 @@
...
+--  linux: ~/.wireshark/plugins/
+--
+-- Note that this uses heuristic dissectors to identify the
+-- fpga packets. Heuristic dissectors were first exposed to
+-- the lua API in version 1.11.3. On some linux distributions,
+-- the packaged version of wireshark is too old, so you may
+-- have to build from source.
+--
+-- For details of the wireshark lua API, see:
+-- https://www.wireshark.org/docs/wsdg_html_chunked/lua_module_Proto.html
```
DanielSank: `https://www.wireshark.org/docs/wsdg_html_chunked/wsluarm_modules.html` instead.

fpgalib/tools/ghz_fpgas.lua
```diff
@@ -0,0 +1,206 @@
...
+
+local p_dac = Proto("ghz_dac", "GHz DAC")
+local f_dac_dest = ProtoField.ether("ghz_dac.dest", "Destination")
```
DanielSank: What's the rationale for the abbreviated and full names? I guess this has something to do with how Wireshark displays and sorts packets...?

maffoo: yeah, the first string (`ghz_dac.dest`) is sort of an internal thing as far as I can tell, while the second string (`Destination`) shows up in the tree view when expanding the dissected packet tree. It also seems that the first string must be unique across all fields, and that a given field object can only be added to one protocol. This is annoying because the ethernet source and destination fields clearly ought to work for both DAC and ADC packets, but wireshark doesn't allow that AFAICT.

DanielSank:
A little bit of narration might help (the next is a bit wall-ish) but it's simple enough that I'm not super sure it's needed.

LGTM.

Issue #276, opened 2015-12-01
FPGA server cannot find ADCs or DACs


Coxious:
Hi,

I'm a student currently studying in Haohua's group, Zhejiang University, China. Recently we want to upgrade our LabRad system to latest version on GitHub, and we meet some trouble here.

We deployed the latest LabRad system, and it cannot find our ADCs and DACs. Calling the list device function can properly print those boards on screen. However when I select an ADC or DAC boards, it shows that it's not a DAC or ADC board. 

We cloned the entire server project and pylabrad project from github.

https://github.com/martinisgroup/servers
https://github.com/maffoo/pylabrad

We run the Delphi version of Direct Ethernet ,1.1.1.

 https://github.com/martinisgroup/servers/blob/master/ghz_fpga_server.py : 1062
```python
    # Allow selecting different kinds of devices in each context.
    def selectedDAC(self, context):
        dev = self.selectedDevice(context)
        if not isinstance(dev, dac.DAC):
            raise Exception('selected device is not a DAC board')
        return dev
```
 I printed the results of  `dev`, and it's a `DeviceWrapper` instance rather than `DAC` instance. Although `DAC` is a subclass of `DeviceWrapper`, but I looked around and only found that the instance is constructed like this:

https://github.com/maffoo/pylabrad/blob/master/labrad/devices.py : 101

```python

class DeviceServer(LabradServer):
    """A server for devices.
    
    Creates a DeviceWrapper for each device it finds, based on a
    user-provided function.  Provides standard settings for listing
    devices, selecting a device for the current context, and
    refreshing the list of devices.  Also, provides for device-locking
    with timeouts.
    """
    name = 'Generic Device Server'
    deviceWrapper = DeviceWrapper
```

and 

https://github.com/maffoo/pylabrad/blob/master/labrad/devices.py : 163

```python
        # start additions
        for name, args, kw in additions:
            if name in self.device_guids:
                # we've seen this device before
                # so we'll reuse the old guid
                guid = self.device_guids[name]
            else:
                guid = self.device_guids[name] = self._next_guid
                self._next_guid += 1
                
            dev = self.deviceWrapper(guid, name)
            yield dev.connect(*args, **kw)
            self.devices[guid, name] = dev
```
Where the `dev` was constructed.

I didn't find where or how did the devices initialized with class ADC or DAC.

DanielSank:
> We cloned the entire server project and pylabrad project from github.
>
> https://github.com/martinisgroup/servers
[...]

PR #282, opened 2015-12-04
Add setting to read all packets until timeout expires.

maffoo:
Discussed in #276 that this could be used to simplify the board autodetection logic.


maffoo:
Note that this would be the first API incompatibility between the old delphi and new scala direct ethernet servers. So if we change the ghz_fpga_server to use this, there'd be no going back.

jwenner:
With the API incompatibility, there should be either be a note in the docstring (both internal and external, if different) or a scala warning (if possible) warning users about the incompatibility. This is to ensure that someone in the future doesn't use this function and accidentally break Delphi compatibility.

Further, at such a time, it might be appropriate to replace the Delphi Direct Ethernet server with something that just raises an Exception saying that the server is deprecated and to use the scala direct ethernet server instead.

maffoo:
@jwenner, I added a note in the docstring (which will be visible to labrad clients) about when the new setting was added.

There's not really a way to deprecate a server; I think our best bet will be to just remove the Delphi binary from the repo. To my knowledge all of our setups have moved to the scala server and no one has used the old delphi version in quite some time.

DirectEthernet/src/main/scala/org/labrad/ethernet/server/DirectEthernet.scala
```diff
@@ -180,6 +180,17 @@ extends ServerContext with Logging {
     }
   }
 
+  private def processPackets(packets: Seq[EthernetPacket]): Seq[(String, String, Int, Array[Byte])] = {
```
DanielSank: What's our stand on docs?

maffoo: docs are good, but in my mind the name and type signature here is really all you need to know. More text here would seem to me to just repeat what's already pretty evident in the code.

DirectEthernet/src/main/scala/org/labrad/ethernet/server/DirectEthernet.scala
```diff
@@ -190,15 +201,7 @@ extends ServerContext with Logging {
...
-    val results = queue.take(numPackets.toInt)
-    results.map { pkt =>
-      val src = pkt.getHeader.getSrcAddr
-      val dst = pkt.getHeader.getDstAddr
-      val typ = pkt.getHeader.getType.value.toInt
-      val data = pkt.getPayload.getRawData
-
-      (src.toString.toUpperCase, dst.toString.toUpperCase, typ, data)
-    }
+    processPackets(queue.take(numPackets.toInt))
```
DanielSank: Isn't converting Long to Int kinda janky?

jwenner:
For deprecation, shouldn't we be able to replace the Delphi direct ethernet exe with a python server which identifies itself as "Direct Ethernet" (just like the Delphi version)? Then have only initServer, where we would raise an Exception with a message about the server being deprecated and to use the scala server instead.

maffoo:
@jwenner, yes, we could do something like that. I just don't think it's worth it. Anyone who needs the old delphi version can get it from the repo history. The rest of us can move on.

DirectEthernet/src/main/scala/org/labrad/ethernet/server/DirectEthernet.scala
```diff
@@ -228,6 +231,24 @@ extends ServerContext with Logging {
...
+            |
+            |Unlike Collect, Read and Discard, this is useful in cases where we
+            |do not know how many packets are expected, but just wish to capture
+            |all traffic in a given interval.
+            |
+            |New in version 1.3.0; not present in Delphi direct ethernet server.""")
+  def readUntilTimeout(): Seq[(String, String, Int, Array[Byte])] = {
+    val timeout = readTimeout.getOrElse { fail(11, "Read Timeout not set") }
+    val millis = timeout.toMillis
+    val nanos = (timeout - millis.milliseconds).toNanos.toInt
```
DanielSank: Wat?

DirectEthernet/src/main/scala/org/labrad/ethernet/server/DirectEthernet.scala
maffoo: If you are asking for more that 2 billion packets, you have bigger problems. We could add a check, but I'm not too worried about it. Using `Long` as a proxy for `unsigned 32-bit int` is dumb, but it's going to take a lot of work to fix that.

jwenner:
@maffoo, the reason would be because other groups seem to be using the FPGA server (and thus I would assume the Direct Ethernet server), so such a deprecation would be a notice to such external users.

DirectEthernet/src/main/scala/org/labrad/ethernet/server/DirectEthernet.scala
```diff
@@ -228,6 +231,24 @@ extends ServerContext with Logging {
   }
 
   @Setting(
+    id = 53,
+    name = "Read until timeout",
+    doc = """Collect and return all packets that arrive until the timeout expires
+            |
+            |Unlike Collect, Read and Discard, this is useful in cases where we
```
DanielSank: I think we need a good ol' Oxford comma after `Read`.

DirectEthernet/src/main/scala/org/labrad/ethernet/server/DirectEthernet.scala
DanielSank: Give the possibility for weird behavior, I'd add the check.

DanielSank:
> @jwenner, yes, we could do something like that. I just don't think it's worth it. Anyone who needs the old delphi version can get it from the repo history. The rest of us can move on.

Agreed. If people need the old direct ethernet server it's because they're using the old fpga server. If people want to stick with and/or fork from old code, that's entirely their decision, and not one we have any reason to support.

@maffoo LGTM but it would be a good idea to check that this works on hardware. Is that what you're doing on redheatdos?

maffoo:
@DanielSank; I was checking the new pcap4j library on redheatdos. We're not yet using the new setting added here, but I could modify the fpga server to use it and test that if you want to see it working.

DanielSank:
Unfortunately I'm away and will be travelling a lot so it's a bit hard for me to test stuff (although I guess you're used to this workflow so maybe you have wisdom). In any case, what would you say to doing the FPGA work that uses this in the same branch and PRing that stuff all-together? At least that way we get to see that it works right.

Alternatively we could set up a real unit test.

PR #293, opened 2015-12-11
Fix asynchrony bugs in dc_rack_server.py

amopremcak:
@DanielSank Thanks for taking the time to write up these notes. I followed your advice and separated the bug fixes from the new features. I must have added and subtracted a bunch of blank lines from this file as I modified it to suit our needs. In any case, I should have payed more attention to these changes when I submitted the initial pull request. I have a decent amount of experience with git, but I am fairly new to pull requests. Seems like a nice way of controlling what changes go into your master branch.


DCRack/dc_rack_server.py
```diff
@@ -607,7 +605,7 @@ def state(self):
 
     def strState(self):
         """Returns the channel state as a list of strings"""
-        s = list(self.state)
+        s = list(self.state()) # self.state is a method, not an attribute...
```
DanielSank: Remove the comment.

DCRack/dc_rack_server.py
```diff
@@ -273,7 +272,6 @@ def changeLEDs(self, data):
             data = 4*data[0] + 2*data[1] + 1*data[2]
         else:
             data &= 0x7
-        self.write([1L])
```
DanielSank: It would be nice if someone could comment on why this was here in the first place. @zchen088 @thchwhite?

DanielSank:
The changes all look good except for that `write([1L])` which has been removed. It looks quite deliberate so I want to know why it was there before we get rid of it.

DCRack/dc_rack_server.py
zchen088: I have no idea what this does.

amopremcak:
I added the changes to the fix_dc_rack_server branch. Sorry for the delay, I've been caught up with finals.

DanielSank:
I say LGTM. However, I think we should get one from @zchen088 too since this is important and I think he's the real expert with this part of the system.

PR #294, opened 2015-12-11
Clean up names in qubit server to follow scala conventions

maffoo:
Includes a bunch of small renaming changes. These are changes of the "if it compiles, it works" sort. Just trying to clean up the naming and remove getters where possible.


pomalley:
Couple questions, LGTM.

PR #296, opened 2015-12-12
Fix rdelay documentation

ejeffrey:
triggering ends when rdelay[15..0] =0

@danielsank please make sure this is correct.


DanielSank:
@ejeffrey how to check?

PR #308, opened 2016-01-23
Add sip.h workaround for pyqt installation

DanielSank:
If your virtualenv has symlinked the `/include/python2.7` *directory*, then the configure and make scripts are unable to access `sip.h` unless you sudo, which leads to other problems. We add a workaround to avoid all this.

Note that we still have a problem where we have to manually change the ownership of a few files. This actually only came up on my goobuntu machine and not elsewhere. I think a good solution for all of this would be to set an installation prefix for the binaries as well so we *never* have to sudo and therefore we wouldn't have to manually change file ownership. I'm actually willing to add that to this PR if someone would volunteer to make a virtualenv and confirm that my instructions work.

@thchwhite


DanielSank:
I now find that after doing this, my virtualenv cannot import `PyQt4.uic`. Not sure why yet.

DanielSank:
Ah this is because when I did `sudo make install` for PyQt4 it set the ownerships wrong.

~Sigh~

DanielSank:
Tested to work on a Goobuntu machine.

maffoo:
There might be a simpler alternative for this, namely to use the prepackaged versions of pyqt and pyqwt installed at the system level. If you set up a virtualenv like:
```
$ mkvirtualenv --system-site-packages myenv
```
Then it will be able to use packages installed at the system level, but you can also add packages in the virtualenv itself that will be isolated. Then you can `apt-get install` prebundled packages that are hard to compile (like numpy, scipy, pyqt and pyqwt) at the system level, and still be able to use them in your virtualenv. I think making this our standard recommendation would simplify the process, not to mention greatly speed things up since you don't have to compile these large packages for every virtualenv.

Note that it's also possible to override system packages with newer versions in a virtualenv. For example, I have ipython installed from apt but it's ancient (1.2.1). In a virtualenv with the `--system-site-packages` option, I can run ipython and I get this system version. If I then do
```
$ pip install --upgrade ipython
```
I can install the latest inside my virtualenv, while still being able to access other system packages like, say, numpy.

Also, just for reference, the apt packages for pyqt and pyqwt are `python-qt4` and `python-qwt5-qt4`, respectively.

DanielSank:
I think perhaps setting the install prefix correctly would be another alternative, but I haven't tried it.

Issue #309, opened 2016-01-27
ADC Register Readback setting does not work for ADC Build 7


patzinak:
See issue #301 for more details.

Issue #301, opened 2016-01-09
ADC Build 7 LED1 Status


patzinak:
When an ADC board with the firmware Build 7 is run in a normal regime, should LED1 ("AD clock not locked" according the documentation) be on? All ADC boards that were upgraded to the latest firmware version (two at the moment) shows the same behavior, namely, the LED1 is basically on all the time. We can switch the LED off by running the initialization script a good number of times (roughly 4-30 times) but the LED always goes on when we try to query its state or when we try to run any measurement code, i.e. whenever we read anything back from the board. The quires always return 1. We still can get quite meaningful data from the boards when we are running any tests or measurements but it is not clear whether we should trust this data. Before the upgrade, however, it was always possible to initialize and run the ADC boards (but not query them) having all LED in the off state. Are Build 1 and Build 7 anyhow different when it comes to LEDs? Overall, Build 7 feels much better but the PLL LED undermines the trustworthiness of the data. The up-to-date LabRAD (Scala-based)/Ethernet (Scala-based) and GHz FPGA servers are used.

@amopremcak, @gribeill

patzinak:
Okay, this is how far I was able to get:

The documentation is really confusing but mostly based on AA_documentaitonAD_v6.doc, my understanding now is that LED[0] on corresponds to unlatching of the "FPGA PLL" or "external 1 GHz PLL" (which seemed to be two different things), while LED[1] on corresponds to "bad clocking phase of the AD" converters, the later is somehow also referred to as "GHz PLL" and this is really, really confusing. I can't find the circuit diagram to check whether this makes sense from the hardware design viewpoint.

There is an apparent bug in GHz FPGAs server: setting `adc_register_readback` does not work for ADC Build 7. The actual bug is in https://github.com/martinisgroup/servers/blob/master/fpgalib/adc.py , line 958. The suggested fix is to replace `regs = self.regRun(self.RUN_MODE_REGISTER_READBACK, 0) # 0 reps?` with `regs = self.regRun(self.RUN_MODE_REGISTER_READBACK, {}, 0)`. Ping: @maffoo , @DanielSank 

I also suggest to add `'dClkBits': a[1] >> 1 & 15,` into the corresponding `processReadback` method. See https://github.com/McDermott-Group/servers/commit/eb879bba78795ad3abfae9e8b847d1d5a6a9e23b .
[...]

PR #310, opened 2016-01-27
Fix bug #309

patzinak:
See bug  #309 separated from issue #301.


fpgalib/adc.py
```diff
@@ -1052,7 +1052,8 @@ def processReadback(resp):
         a = np.fromstring(resp, dtype='<u1')
         return {
             'build': a[0],
-            'noPllLatch': a[1]&1 == 1,
+            'noPllLatch': bool(a[1] & 1),
+            'dClkBits': a[1] >> 1 & 0b1111,
```
DanielSank: I would really like it if @jrainbo could verify this.

DanielSank:
This looks great up to a minor comment and the fact that the tests are failing. That might be a problem with the test server though.

fpgalib/adc.py
```diff
@@ -955,7 +955,7 @@ def registerReadback(self):
         @inlineCallbacks
         def func():
             # build registry packet
-            regs = self.regRun(self.RUN_MODE_REGISTER_READBACK, 0) # 0 reps?
+            regs = self.regRun(self.RUN_MODE_REGISTER_READBACK, {}, 0)
```
maffoo: maybe use named params here to help the reader:
```python
regs = self.regRun(self.RUN_MODE_REGISTER_READBACK, info={}, reps=0)
```

fpgalib/adc.py
maffoo: I can never remember the preference of operators. I'd prefer to be explicit:
```python
'dClkBits': (a[1] >> 1) & 0b1111,
```

fpgalib/adc.py
patzinak: Naming of the params is a good idea here but this should be done in a consistent way across the module. Should I add a commit that adds param names `info` and `reps` to **all** `self.regRun()` calls in the module?

maffoo: I think it'd be a good idea to add param names throughout the module. Note that the param names for build1 and build7 are different.

fpgalib/adc.py
patzinak: Done.

fpgalib/adc.py
patzinak: Done, this should be reviewed.

fpgalib/adc.py
```diff
@@ -441,9 +439,10 @@ def runAverage(self, filterFunc, filterStretchLen, filterStretchAt,
         @inlineCallbacks
         def func():
             # build registry packet
-            regs = self.regRun(self, self.RUN_MODE_AVERAGE_AUTO, 1, filterFunc,
-                filterStretchLen, filterStretchAt, demods)
-            
+            regs = self.regRun(self, mode=self.RUN_MODE_AVERAGE_AUTO,
```
maffoo: Well, this is just wrong. We're calling `regRun`, a classmethod, through an instance of that class, and then passing an unneccessary self reference as the first argument. I realize you didn't create the issue here, but it stands out that this needs to be fixed. I'll file a separate issue.

maffoo: #312

maffoo:
LGTM, but @DanielSank do you still want to have @jrainbo look at it?

DanielSank:
It would be really nice if he could look at the circuit diagram and just check that the dclk bits variable here agrees. I say this based on the following excerpt from #301

> My understanding is that when the board operates in a normal regime all four bits should report 1. While the exact meaning of these AD bits is somewhat obscure this seems like an extremely useful diagnostic feature, which helped me, and hopefully we can claim this now, partially resolve the original issue.

patzinak:
I'd say that I'm sure that these bits carry useful information. This is according to _documentationGHzAD_v4.doc and due to the fact that I saw them being perfectly correlated with the LED[1] status. The documentation says:
```
d(2)        clockmon[7..0]        bit0 = NOT(LED1 light) = no PLL in external 1GHz clock
                                  bit1=dclkA output (phase of data stream)
                                  bit2=dclkA delayed 1ns output
                                  bit3=dclkB output
                                  bit4=dclkB delayed 1ns output
```
The documentation, however, does not tell how to interpret these bits (unlike in the case with `bit0`). That's why I'm returning a number, not four booleans (like `unlockedClkA` and so on).

I had to come up with an interpretation and I believe that all bits should return ones when everything works correctly, however, this assumption/guess is not reflected in the code in any way because this is not documented.

DanielSank:
> The documentation, however, does not tell how to interpret these bits

Yes, exactly. This is the right time to just figure it out and get it right here in the code. @jrainbo

fpgalib/adc.py
patzinak: I see quite a few things in the code that does not really look great to me. However, I keep in mind the following: (1) I'm not 100% familiar with the code, there might be tricky, uncommented bug fixes, not obviously standing out in the commit history; (2) the code works for your group; (3) we would love to see our core code synced with yours to benefit from all future updates. It seems that the right way for us to act is to ask for the minimum number of changes that should do the job for us.

DanielSank: Please post issues. It's perfectly fine to post an issue just asking why a bit of code is a certain way (we even have a "question" label for issues).

maffoo: I think I was being a bit overdramatic about classmethods. We should just remove the first `self` argument to the method. If you can do that here while you're touching these lines, @patzinak, that'd be great.

patzinak:
Sure, that would be great!

Is @jrainbo getting the notifications?

DanielSank:
Not sure. I may have to try notification via air phonons.

jwenner:
It might be good to bump the version number at https://github.com/martinisgroup/servers/blob/master/ghz_fpga_server.py#L180.

fpgalib/adc.py
patzinak: This has been addressed in a corresponding commit.

PR #316, opened 2016-02-07
Added some instructions and an optional password for authentication p…

amopremcak:
…urposes.


Telecomm/telecommServer.py
```diff
@@ -127,6 +151,7 @@ def _refreshConnectionData(self):
         reg = cxn.registry
         p = reg.packet()
         p.cd(REGISTRY_PATH)
+        p.dir(key = 'regcontent')
```
DanielSank: No space here, i.e. `p.dir(key='regcontent')`

DanielSank:
To be honest, I don't really understand good security practices. @maffoo is it normal to store a password in memory as is done here?

amopremcak:
I removed the blank space. I tested it for backwards compatibility. Also, how do you handle your version numbers? Should I have bumped this from 1.1 to 1.2?

DanielSank:
Good question. It's backwards compatible, so bump the minor version number.

amopremcak:
Done.

amopremcak:
If it turns out you guys think it is better to remove the password from the registry, I can easily change this. I was sort of wondering the same thing you asked @maffoo but for our purposes, the email account being used is really just a burner account.

PR #324, opened 2016-02-25
Fix unwanted exceptions from dac 15 class.

### LINKED ISSUE #323: dac.py raises jump table error incorrectly
[This line](https://github.com/martinisgroup/servers/blob/master/fpgalib/dac.py#L1489) raises an error about ```from_addr``` being too close together on a valid sequence.

The FPGAs have a restriction that jump table op codes cannot be executed more often than every 4 clock cycles, due to pipelining. However, this code is looking for how ```from_addr``` are defined __in memory__, not at what times they are executed. 

I believe the only way to fix this is to emulate the experiment and look for when this may be violated, unless I am mistaken. 

I think the best course of action may be to remove this check altogether for now. Or it might be better to remove this check and move it into pyle for the time being, where we already emulate the jump table to make the retrigger table.

also ```JT_MIN_FROM_ADDR_SPACING = 2``` when it should ```=4```

JulianSKelly:
Fixes #323

Change error checking for when two jump table commands are executed too closely in time, rather than too closely in memory.

### DIFF
--- a/fpgalib/dac.py
+++ b/fpgalib/dac.py
@@ -1172,6 +1172,23 @@ class DAC_Build15(DAC_Build8):
     19  d[1]
     ...
     33  d[15]
+
+    Table documenting how each op code can move the sram, jump table index
+
+    OP    | BEHAVIOR
+    ------------------------------------------------------------------------
+    IDLE  | increment sram_idx and increment jt_idx
+    NOP   | increment sram_idx and increment jt_idx
+    CHECK | move sram_idx and jt_idx OR increment sram_idx and jt_idx
+    JUMP  | move sram_idx and jt_idx
+    CYCLE | move sram_idx and jt_idx OR increment sram_idx and jt_idx
+    END   | no next entry
+
+    Attributes:
+        increment_ops (list[jump_table.Operation]): Jump table op codes that can
+            increment the sram index, jump table index
+        jump_ops (list[jump_table.Operation]): Jump table op codes that can
+            jump the sram index, jump table index
     """
 
     HAS_JUMP_TABLE = True
@@ -1181,21 +1198,20 @@ class DAC_Build15(DAC_Build8):
     SRAM_WRITE_PKT_LEN = 256
     SRAM_WRITE_DERPS = SRAM_LEN // SRAM_WRITE_PKT_LEN
 
-    JUMP_TABLE_LEN = 528
+    JUMP_TABLE_PACKET_LEN = 528
     NUM_COUNTERS = 4
     COUNTER_BYTES = 4
     IDLE_BITS = 15
     JUMP_TABLE_ENTRY_BYTES = 8
-    JUMP_TABLE_COUNT = (JUMP_TABLE_LEN - (NUM_COUNTERS * COUNTER_BYTES)
-                        ) / JUMP_TABLE_ENTRY_BYTES
+    JUMP_TABLE_COUNT = 64
     JT_FROM_ADDR_OFFSET = -2
     JT_END_ADDR_OFFSET = JT_FROM_ADDR_OFFSET - 1
     JT_IDLE_OFFSET = -0
     JT_IDLE_MIN = 0 - JT_IDLE_OFFSET
     JT_IDLE_MAX = 2**IDLE_BITS - 1 - JT_IDLE_OFFSET
-    JT_MIN_FROM_ADDR_SPACING = 2
-    JT_MIN_FROM_ADDR = JT_MIN_FROM_ADDR_SPACING - JT_FROM_ADDR_OFFSET
-    JT_MIN_END_ADDR = JT_MIN_FROM_ADDR_SPACING - JT_END_ADDR_OFFSET
+    JT_MIN_CLK_CYCLES_BETWEEN_OPCODES = 4
+    JT_MIN_FROM_ADDR = JT_MIN_CLK_CYCLES_BETWEEN_OPCODES
+    JT_MIN_END_ADDR = JT_MIN_CLK_CYCLES_BETWEEN_OPCODES
     JT_MIN_TO_ADDR = 0
     JT_MAX_FROM_ADDR = SRAM_LEN // 4 + JT_FROM_ADDR_OFFSET
     JT_MAX_END_ADDR = SRAM_LEN // 4 + JT_END_ADDR_OFFSET
@@ -1204,6 +1220,15 @@ class DAC_Build15(DAC_Build8):
 
     MONITOR_0 = 5
     MONITOR_1 = 10
+
+    increment_ops = (
+        jump_table.IDLE, jump_table.NOP, jump_table.CHECK,
+        jump_table.CYCLE
+    )
+
+    jump_ops = (
+        jump_table.CHECK, jump_table.JUMP, jump_table.CYCLE
+    )
 
     @classmethod
     def convert_from_address(cls, x_ns):
@@ -1430,7 +1455,7 @@ class DAC_Build15(DAC_Build8):
         :param str name: one of: END, NOP, IDLE, CYCLE, JUMP, CHECK, RAMP
         :param list[int] arg: arguments for the op.
         :return: list of jump table entries
-        :rtype: list[jump_table.JumpEntry]
+        :rtype: jump_table.JumpEntry
         """
         if name == 'CHECK':
             raise NotImplementedError("Check not implemented yet")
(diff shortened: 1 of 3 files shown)


fpgalib/dac.py
```diff
@@ -1430,7 +1430,7 @@ def make_jump_table_entry(cls, name, arg):
         :param str name: one of: END, NOP, IDLE, CYCLE, JUMP, CHECK, RAMP
         :param list[int] arg: arguments for the op.
         :return: list of jump table entries
-        :rtype: list[jump_table.JumpEntry]
+        :rtype: jump_table.JumpEntry
```
pomalley: huh, good catch here.

fpgalib/dac.py
```diff
@@ -1477,19 +1477,56 @@ def make_jump_table_entry(cls, name, arg):
...
+        current command, arguments and conditions. We document all commands
+        below.
+
+        OP    | BEHAVIOR
+        IDLE  | increment sram_idx and increment jt_idx
+        NOP   | increment sram_idx and increment jt_idx
+        CHECK | move sram_idx and jt_idx OR increment sram_idx and jt_idx
+        JUMP  | move sram_idx and jt_idx
+        CYCLE | move sram_idx and jt_idx OR increment sram_idx and jt_idx
+        END   | no next entry
```
pomalley: I would recommend
```
OP    | BEHAVIOR
----------------  # <-- add horizontal rule
IDLE  | ...
```

fpgalib/dac.py
```diff
@@ -1477,19 +1477,56 @@ def make_jump_table_entry(cls, name, arg):
...
+        # these entries can increment both idx
+        increment_idx = (
+            jump_table.IDLE, jump_table.NOP, jump_table.CHECK,
+            jump_table.CYCLE
+        )
+
+        # these entries can move both idx
+        jump_idx = (
+            jump_table.CHECK, jump_table.JUMP, jump_table.CYCLE
+        )
```
maffoo: since these lists are constant, I would build them outside the function:
```python
# ops that can increment the JT and SRAM indices
IncrementOps = (
    jump_table.IDLE,
    jump_table.NOP,
    ...
)

# ops that can jump the JT and SRAM indices
JumpOps = (
    jump_table.CHECK,
    jump_table.JUMP,
    ...
)
```

Also, because you can pass a tuple to `isinstance` you can use these in the for loop when you check the types:
```python
for k, jt_entry in enumerate(jt_entries):
    if isinstance(jt_entry.operation, IncrementOps):
        ...
    if isinstance(jt_entry.operation, JumpOps):
        ...
```

fpgalib/dac.py
```diff
@@ -1477,19 +1477,56 @@ def make_jump_table_entry(cls, name, arg):
...
+
+        # verify no two from_addr are executed within JT_MIN_FROM_ADDR_SPACING
```
pomalley: update comment: JT_MIN_FROM_ADDR_SPACING doesn't exist anymore

fpgalib/dac.py
```diff
@@ -1477,19 +1477,56 @@ def make_jump_table_entry(cls, name, arg):
...
+        for k, jt_entry in enumerate(jt_entries):
+            # first pass we check all increment cases
+            if type(jt_entry.operation) in increment_idx:
+                sram_idx = jt_entry.from_addr
+                next_jt_idx = k + 1
+                next_from_addr = jt_entries[next_jt_idx].from_addr
+                assert ((next_from_addr - sram_idx) >
+                        cls.JT_MIN_CLK_CYCLES_BETWEEN_OPCODES)
```
pomalley: Might be good to add a message here: `assert condition, "Entries {} and {} are too close"` or something.

Might also want to make a custom exception type, I dunno.

fpgalib/dac.py
```diff
@@ -1477,19 +1477,56 @@ def make_jump_table_entry(cls, name, arg):
...
+            jump_table.CHECK, jump_table.JUMP, jump_table.CYCLE
+        )
+
+        # verify no two from_addr are executed within JT_MIN_FROM_ADDR_SPACING
+        for k, jt_entry in enumerate(jt_entries):
+            # first pass we check all increment cases
+            if type(jt_entry.operation) in increment_idx:
+                sram_idx = jt_entry.from_addr
+                next_jt_idx = k + 1
+                next_from_addr = jt_entries[next_jt_idx].from_addr
```
pomalley: Do we check anywhere that the last entry is an END? Or does that get added automatically (probably)? If not it will error here with a confusing message.

fpgalib/dac.py
```diff
@@ -1477,19 +1477,56 @@ def make_jump_table_entry(cls, name, arg):
...
+                assert ((next_from_addr - sram_idx) >
+                        cls.JT_MIN_CLK_CYCLES_BETWEEN_OPCODES)
+            # second pass we check all jump cases
+            if type(jt_entry.operation) in jump_idx:
+                sram_idx = jt_entry.to_addr
+                # take off 1 as first entry is always NOP (self.toString)
+                next_jt_idx = jt_entry.operation.jump_index - 1
```
maffoo: Why is there a -1 here? I thought the NOP would be added already and the indices fixed up before sending anything to the fpga server.

Also, what is the `self.toString` supposed to indicate?

maffoo: ok, I see now that the `toString` method implicitly adds a NOP in addition to the listed entries. This is really unfortunate, because it means the jump table entries we create are not self-consistent, since the numerical indices don't match up with the indices in the list of entries. I'd much rather do either of the following:
1. require the user to add the NOP, so that then the relative jt indices are consistent (and identical to what goes in fpga memory)
2. OR, add  the NOP as here but also add one to jump indices so that when we compute jt indices they point to the right places in our list.

I would lean toward 2. Thoughts?

fpgalib/jump_table.py
```diff
@@ -27,6 +27,8 @@ def __init__(self, from_addr, to_addr, operation):
         :param int to_addr: the to address (should be 0 if not used)
         :param Operation operation: the operation
         """
+
+        # TODO: Why is this -1 hard coded?
```
maffoo: AFAICT from the docs, there is no restriction on the relationship between SRAM addresses `from_addr` and `to_addr` on a single jump table op. I think this check is just bogus.

fpgalib/dac.py
DanielSank: Looking at the code, I don't see where this supposed NOP is inserted in `toString`.

maffoo: You have to squint very hard, but [this line](https://github.com/martinisgroup/servers/blob/u/julian/test-dac/fpgalib/jump_table.py#L331) and the previous line puts op code 5 (NOP) as the first entry in the jump table. This is required, it's just a question of where we enforce that requirement, and how to do it in a way that the other jump table calculations are self-consistent.

DanielSank: Oh jeez. That was really stupid, sorry. I like option 1 over option 2, but I realize it's more work.

pomalley: For the record, the reason that it ended up this way is because the first JT op code is required to be NOP, and the argument to it is the start address. Since we already had parameters being passed around for the start address from the previous build of the DACs, it seemed natural to just use those and stick it in there when constructing the JT packet. Perhaps not ideal given that it essentially 1-indexes the rest of the JT entries, but there you have it.

maffoo: Given that this is a larger problem that would require coordinated changes across server and client software, I think it's best not to let it hold up the rest of this PR. I'll file a separate issue.

pomalley: Yes, I was going to suggest that as well. Thanks.

maffoo: Filed #325

fpgalib/test/test_dac.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def test_catch_bad_jump_table(self):
+        """Test jump table commands that are executed too frequently"""
+
+        # from_addr of 16 followed by 20 are too close together
+        bad_jump_args = [
+            ('NOP', [16]),
+            ('NOP', [20]),
+            ('END', [400])
+        ]
```
maffoo: I think it's best here to split these cases into separate test methods and give them descriptive names rather than having multiple cases within one methods with comments to explain what's different. I also generally think that if test methods are small, test one thing, and named descriptively, then docstrings are usually not needed on each method.
```python
def test_nop_nop_from_addr_too_close(self):
    entries = self.make_entries([
        ('NOP', [16]),
        ('NOP', [20]),
        ('END', [400])
    ])
    with pytest.raises(AssertionError):
        self.dac.make_jump_table(entries)

def test_jump_nop_from_addr_too_close(self):
    ...
```
Obviously this slightly increases the amount of boilerplate, but I think it makes debugging test failures much easier.

fpgalib/test/test_dac.py
```diff
@@ -0,0 +1,95 @@
...
+                     self.dac.JT_MIN_CLK_CYCLES_BETWEEN_OPCODES * 4]),
+            ('END', [400])
+        ]
+
+        entries = self.make_entries(bad_jump_args)
+        with pytest.raises(AssertionError):
+                self.dac.make_jump_table(entries)
+
+        from_addr = self.dac.JT_MIN_FROM_ADDR * 4
+        bad_jump_args = [
```
maffoo: Again, I would suggest splitting up the good and bad cases into separate, smaller test methods. Also, this table of entries should not be named `bad_*` since it represents a valid jump table.

fpgalib/dac.py
JulianSKelly: Cool! Didn't know this.

JulianSKelly: Do you suggest make these class level attributes at the top?

fpgalib/dac.py
JulianSKelly: AFAIK we don't check for this

JulianSKelly: It definitely doesn't get added automatically as the user needs to specify when a sequence will end.

fpgalib/dac.py
JulianSKelly: Yeah this was super confusing, it's actually not well described in the documentation that all sequences must begin with an NOP. I happened to remember this was the case but it still took me some time to find it.

fpgalib/jump_table.py
JulianSKelly: Yeah I thought the same thing but I am not 100% sure

fpgalib/dac.py
maffoo: Might be best to make these module-level constants in `jump_table.py`. And yes, I would move the documentation there as well.

JulianSKelly:
@maffoo @pomalley I believe I have addressed everything

maffoo:
A couple of comments, but otherwise LGTM.

JulianSKelly:
Addressed your comments. Will merge after I can test on hardware to make sure nothing funny happened.

JulianSKelly:
@maffoo Checked this on hardware and found a missing error for jump tables that are too long.

Can you inspect latest commit and LGTM if ready?

JulianSKelly:
@maffoo please see my changes.

JulianSKelly:
Also looks like teamcity is barfing on futures

JulianSKelly:
@maffoo, can you take another look?

maffoo:
LGTM

PR #327, opened 2016-02-28
Add test for last jt entry == END.

### LINKED ISSUE #326: There is no check that jump tables end with END()

JulianSKelly:
fixes #326


pomalley:
Heh, well, the exception caught another test where we don't add an END.

JulianSKelly:
Closing this as some calibrations require continuous looping without ```END```

PR #336, opened 2016-03-31
Refactor, move fpga test for jump table in/to servers.

JulianSKelly:
Improve DRY-ness by having this module _only_ interface
with the boards through the ghz_fpga_server.

Write basic hardware testing functions that exercise all
jump table commands, have them sensitive to correct
firmware offsets.

*do_end
*do_nop
*do_idle
*do_jump
*do_cycle


fpgalib/dac.py
```diff
@@ -83,6 +83,46 @@
 # TODO
 # Think of better variable names than self.params and self.boardParams
 
+# TODO: Integrate this with DAC class
+# TODO: Have dac_calibration_server not DACify SRAM data
+def waves2sram(wave_a, wave_b, trigger_idx=(16, 17)):
```
DanielSank: I'd rather call this `dacify`, and in either case we should take this opportunity to fix the name (i.e. no camelCase).

JulianSKelly: fixed

DanielSank:
LGTM

PR #337, opened 2016-04-01
Test dacify and reduce loops through waveforms.

DanielSank:
@JulianSKelly

### DIFF
--- a/fpgalib/dac.py
+++ b/fpgalib/dac.py
@@ -85,41 +85,43 @@ import fpgalib.jump_table as jump_table
 
 # TODO: Integrate this with DAC class
 # TODO: Have dac_calibration_server not DACify SRAM data
-def dacify(wave_a, wave_b, trigger_idx=(16, 17)):
-    """Construct SRAM sequence for a list of waveforms.
+
+
+SRAM_MAX = 0x1FFF  # 8190 = 2**13 - 1
+
+
+def dacify(wave_a, wave_b, trigger_idx=None):
+    """Construct SRAM sequence for a list of waveforms (DACify waveforms).
 
     Warning: Lots of hardware details exposed here.
 
     Args:
-        wave_a (np.ndarray): Waveform for DAC A normalized to ???
-        wave_b (np.ndarray): Waveform for DAC B normalized to ???
+        wave_a (np.ndarray): Waveform for DAC A. Maximum amplitude is 1, i.e.
+            values ranging from -1 to 1 indicate full scale.
+        wave_b (np.ndarray): Same as previous, but for DAC B.
 
-    Returns (np.ndarray): Array of SRAM data with shape ???
+    Returns:
+        (list): Each element is an integer representing a 32 bit SRAM word.
     """
-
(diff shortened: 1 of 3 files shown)


fpgalib/dac.py
```diff
@@ -85,41 +85,43 @@
...
+
+def dacify(wave_a, wave_b, trigger_idx=None):
+    """Construct SRAM sequence for a list of waveforms (DACify waveforms).
 
     Warning: Lots of hardware details exposed here.
 
     Args:
-        wave_a (np.ndarray): Waveform for DAC A normalized to ???
-        wave_b (np.ndarray): Waveform for DAC B normalized to ???
+        wave_a (np.ndarray):
```
JulianSKelly: Blank arg

fpgalib/test/test_dac.py
```diff
@@ -5,6 +5,21 @@
 import fpgalib.dac as dac
 
 
+def test_waves2sram():
```
JulianSKelly: Name should be test_dacify

fpgalib/dac.py
```diff
@@ -85,41 +85,43 @@
...
 # TODO: Integrate this with DAC class
 # TODO: Have dac_calibration_server not DACify SRAM data
-def dacify(wave_a, wave_b, trigger_idx=(16, 17)):
-    """Construct SRAM sequence for a list of waveforms.
+
+
+SRAM_MAX = 0x1FFF  # 8190 = 2**13 - 1
+
+
+def dacify(wave_a, wave_b, trigger_idx=None):
```
JulianSKelly: Could you change the default triggers in fpgatester to compensate for default value removal?

fpgalib/dac.py
```diff
@@ -85,41 +85,43 @@
...
-    # Chop off everything except lowest 14 bits.
-    trunc_a = [y & 0x3FFF for y in data_a]
-    trunc_b = [y & 0x3FFF for y in data_b]
-    dac_a_dat = trunc_a
-    # Shift DAC B data by 14 bits.
-    dac_b_dat = [y << 14 for y in trunc_b]
+
+    # Multiply wave by full scale. Then truncate to lowest 14 bits. Then for dac
+    # B shift left by 14 bits.
+    dac_a_dat = [0x3FFF & (long(np.floor(SRAM_MAX*y))) for y in wave_a]
```
JulianSKelly: Why do we have to truncate when we verified it doesn't exceed +/-1? Are you truncating the bits below 14 bit resolution? Should the comment say truncate to largest 14 bits?

JulianSKelly:
Minor comments and LGTM. Please change fpgatester trigger idx default.

fpgalib/dac.py
DanielSank: The truncation might not be necessary, but it was there before and I don't want to remove it without thinking carefully. As for the comment, it's definitely keeping the lowest bits, not the largest.

DanielSank:
@JulianSKelly see latest commit

PR #338, opened 2016-04-01
Add explicit licenses for direct ethernet and qubit servers.

maffoo:
@pomalley, @DanielSank


pomalley:
LGTM, given the discussion

PR #335, opened 2016-03-21
Update multi-headed data vault to work with pylabrad 0.95+

maffoo:
We refactor the DataVaultConnector to be a Service instead of a MultiService, and to use a single inlineCallbacks coroutine to manage the reconnection logic for a given manager. This lets us isolate the server start and stop logic which changed with pylabrad 0.95 so that we can handle both the old and new pylabrad versions.

PR #340, opened 2016-04-18
Fixed a unit error in ADR server PID control

andrewdunsworth:


DanielSank:
Thanks, @andrewdunsworth. Can you add the unit labels to the strings? So, instead of seeing `T: 4` we'd see `T: 4 K`.

zchen088:
@DanielSank do the severs tests usually pass? They seem unrelated to the change Andrew made.

DanielSank:
@zchen088 No idea. Check the error on TeamCity?

andrewdunsworth:
Done.
What check did I fail here?

agilent_3640A_dc_source.py
```diff
@@ -78,7 +78,7 @@ def current(self, c, curr=None):
         """
         dev = self.selectedDevice(c)
         if not dev.psMode and curr is not None:
-            yield dev.write('CURR %g' % float(curr))
+            yield dev.write('CURR %g' % float(curr['A']))
```
DanielSank: Units?

andrewdunsworth:
What is dev.write doing?

DanielSank:
`dev.write` tells the GPIB server to send a GPIB command over the wire.

andrewdunsworth:
Better?

DanielSank:
@andrewdunsworth yeah it's better to use `format`. Nice.

I was wondering if you could put the units in the print statements too.

andrewdunsworth:
I think the '{}'.format() will do that.

zchen088:
@DanielSank 

```
In [42]: '{}'.format(1*ns)
Out[42]: '1.0 ns'
```

zchen088:
@maffoo can you provide any insight on the failed tests?

DanielSank:
LGTM but someone should go to TeamCity and see why the tests are failing.

andrewdunsworth:
I don't know what my user name or password is for that? It's not the same as github...

zchen088:
I looked, it seems like the datavault is down for some reason?

jwenner:
@maffoo, could these test failures be due to #335 not being merged in yet?

jwenner:
There are some additional failures about the first argument to servers in `fpgalib.test.test_dac.top_level_collect`, `fpgalib.test.test_fpga_server.top_level_collect`. An example is below:

```
('fpgalib/test/test_dac.py', None, 'fpgalib/test/test_dac.py')
servers/fpgalib/test/test_dac.py:5: in <module>
    import fpgalib.dac as dac
servers/fpgalib/dac.py:29: in <module>
    import fpgalib.fpga as fpga
servers/fpgalib/fpga.py:5: in <module>
    from labrad.devices import DeviceWrapper
pylabrad/labrad/devices.py:91: in <module>
    class DeviceServer(LabradServer):
pylabrad/labrad/devices.py:313: in DeviceServer
    def select_device(self, c, key=0):
pylabrad/labrad/decorators.py:44: in decorator
    return Setting(func, lr_ID, lr_name, returns, unflatten, **params)
pylabrad/labrad/decorators.py:131: in __init__
    "unless other arguments are optional".format(args[0]))
E   ValueError: First argument key cannot accept '' unless other arguments are optional
```

ejeffrey:
Hi Jim,

This is a change in the setting decorator that looks like it has a bug.  I
am looking into it.

ejeffrey:
@jwenner, @maffoo

decorators,py, the if statement on line 129 should be either 'Nrequired >1'
or 'Nrequired != 0'.  Can you guys figure out which one and update it?

pomalley:
@andrewdunsworth how well does the PID work? The fact that this units error is still in there makes me realize just how long it has been since I last even tried the PID functionality...

andrewdunsworth:
It works for what I want it to do now, mag up and down for field cools.

andrewdunsworth:
So anyways, do I merge this now?

jwenner:
I just did a fake PR (#341) which only contains a change to the README in order to determine if the test failures are here or in some more fundamental level. That PR has the same test failures, so my inclination would be to ignore the test failures on this PR but raise them as separate issue(s).

PR #343, opened 2016-04-19
datavault: Fix labrad_urlencode to work with new FlatData type

maffoo:
The FlatData type is a namedtuple that replaces the plain tuple returned by older versions of pylabrad when flattening data. We modify the labrad_urlencode method to use names of the FlatData items if possible, rather than explicitly unpacking the tuple. This lets us maintain compatibility with changes in 0.95.2 which added a third endianness member to the FlatData tuple.

### DIFF
--- a/datavault/backend.py
+++ b/datavault/backend.py
@@ -40,8 +40,14 @@ def time_from_str(s):
     return datetime.datetime.strptime(s, TIME_FORMAT)
 
 def labrad_urlencode(data):
-    data_bytes, t = T.flatten(data)
-    all_bytes, _ = T.flatten((str(t), data_bytes), 'ss')
+    if hasattr(T, 'FlatData'):
+        # pylabrad 0.95+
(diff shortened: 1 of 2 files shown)


jwenner:
FYI, there are still failed datavault tests, although I don't know enough to determine if these relate to #335.

ejeffrey:
test_datavault.py:test_parameters() should have its calls to T.flatten changed to:

```
flat_a = T.flatten(a)
flat_b = T.flatten(b)
assert flat_a == flat_b
```

which will work independent of the number of elements in the FlatData tuple

ejeffrey:
Alternately, assert the individual elements of flat_a and flat_b are equal by name, rather than using tuple unpacking.

maffoo:
@ejeffrey, I've fixed the tests. Could you take another look?

ejeffrey:
LGTM

PR #345, opened 2016-05-02
Make signals class attributes instead of instance attributes

ejeffrey:
Please ignore, I just want to trigger tests on teamcity


maffoo:
The extra commits can be removed now that #343 is in. As for the signal change, I think we cannot move these to the class level with the current signal implementation because of context interference in the multi-headed datavault.

PR #346, opened 2016-05-03
Amj/pyserial version 3 update

jayich:
@ejeffrey , @maffoo , @DanielSank 

With pyserial 3.0. the API changed, http://pythonhosted.org/pyserial/index.html#.

The baudrate method (line 204) runs with a server.  I changed the other attributes according to the new API but they have not been tested with test code, or in a meaningful way with a server.

Feel free to simply use this PR as guidance for making internal changes to serial_server to accommodate the new API.


serial_server.py
```diff
@@ -216,13 +217,13 @@ def baudrate(self, c, data=None):
     def bytesize(self, c, data=None):
         """Sets the bytesize."""
         ser = self.getPort(c)
-        bsizes = [long(x[1]) for x in ser.getSupportedByteSizes()]
+        bsizes = [long(x[1]) for x in ser.BYTESIZES]
```
ejeffrey: This should just be bsizes = ser.BYTESIZES (or just get rid of the temporary altogether)

serial_server.py
```diff
@@ -232,14 +233,14 @@ def bytesize(self, c, data=None):
     def parity(self, c, data=None):
         """Sets the parity."""
         ser = self.getPort(c)
-        bsizes = [x[1] for x in ser.getSupportedParities()]
+        bsizes = [x[1] for x in ser.PARITIES]
         if data is None:
```
ejeffrey: Again, no need for the list comprehension, ser.PARITIES is just a sequence of strings.

serial_server.py
```diff
@@ -249,13 +250,13 @@ def parity(self, c, data=None):
     def stopbits(self, c, data=None):
         """Sets the number of stop bits."""
         ser = self.getPort(c)
-        bsizes = [long(x[1]) for x in ser.getSupportedStopbits()]
+        bsizes = [long(x[1]) for x in ser.STOPBITS]
```
ejeffrey: ditto

ejeffrey:
Hi anderew, thanks this looks great.  I did check, and the pythonic API is supported already it at least pyserial 2.6, and probably earlier so removing the deprecated java style API shouldn't cause any problem with backwards compatibility.

jayich:
@ejeffrey , let me know if you guys want a requirements.txt in the repository.

jayich:
@ejeffrey , any feedback here on my latest commits would be great.

ejeffrey:
I think we don't need the requirements.txt since we don't normally install this repo with pip, but it doesn't hurt.  I would suggest that we document the requirements for each server at the top of the module rather than at the repo level.

ejeffrey:
Otherwise LGTM

maffoo:
I removed requirements.txt as per @ejeffrey's comment and merged this to master. Thanks, @jayich!

jayich:
@maffoo , thank you.  I agree with not having duplication of the pyserial requirement.  I have a feeling you guys have some fancy software distribution scheme, but just in case, make sure everyone pip upgrades pyserial when they next pull master.

PR #350, opened 2016-06-15
Added methods for Setting AC voltage and frequency

btchiaro:
and for clearing the event register.  Tested on Vince.

### DIFF
--- a/agilent_33120a_function_gen.py
+++ b/agilent_33120a_function_gen.py
@@ -44,23 +44,21 @@ class AgilentFunctionGenerator(GPIBManagedServer):
     name = 'Agilent 33120a generator'
     deviceName = 'HEWLETT-PACKARD 33120A'
     
-    @setting(11, 'clear')      
+    @setting(11, 'clear')
     def clear(self, c):
         """Clears status byte summary and event registers
 
         """
-
         dev = self.selectedDevice(c)
         dev.write('*CLS')
       
-    @setting(12, 'set impedance', val='s', returns='')      
-    def set_impedance(self, c, val = '50'):
+    @setting(12, 'set impedance', val='s', returns='')
+    def set_impedance(self, c, val='50'):
         """Sets the loading impedance
         
-        args: 
-           val (str): The loading impedance
+        Args: 
+           val (str): The loading impedance.  Allowed values are '50' or 'INF'
         """
-
         allowed = ['50', 'INF']
         if setting not in allowed:
             raise Exception('allowed settings are: %s' % allowed)
@@ -68,50 +66,43 @@ class AgilentFunctionGenerator(GPIBManagedServer):
         dev.write('OUTP:LOAD {}'.format(val))
 
     @setting(13,'Set DC voltage', val='v[V]', returns='')
-    def set_dc_voltage(self, c,val=0*V):
+    def set_dc_voltage(self, c, val=0*V):
         """Puts generator into DC mode with given voltage.
         
-        args: 
+        Args: 
             val (value[V]): The DC Voltage to be set
-
         """
-        if val< -5*U.V or val > 5*U.V:
+        if val < -5*U.V or val > 5*U.V:
             raise Exception(
-                    'Signal Gnerator only puts out -5 to 5 volts in DC Voltage')
+                    'Signal Generator only puts out -5 to 5 volts in DC Voltage')
         dev = self.selectedDevice(c)
         dev.write('APPL:DC')
         dev.write('VOLT:OFFS {}'.format(val['V']))
     
-    @setting(14, 'set AC voltage', val='v[V]', returns='')      
-    def set_ac_voltage(self, c, val = '1*U.V'):
+    @setting(14, 'set AC voltage', val='v[V]', returns='')
+    def set_ac_voltage(self, c, val='1*U.V'):
         """Puts generator into AC mode with given peak to peak voltage.
         
-        args: 
+        Args: 
             val (value[V]): The AC Voltage to be set
         """
-
         dev = self.selectedDevice(c)
         dev.write('APPL:SIN')
         dev.write('SOUR:VOLT {}'.format(val['V']))
     
-    @setting(15, 'set frequency', val='v[Hz]', returns='')      
-    def set_frequency(self, c, val = '100*U.Hz'):
+    @setting(15, 'set frequency', val='v[Hz]', returns='')
+    def set_frequency(self, c, val='100*U.Hz'):
         """Puts generator into AC mode with given frequency
         
-        args: 
+        Args: 
             val (value[Hz]): The frequency to be set
         """
         dev = self.selectedDevice(c)
         dev.write('APPL:SIN')
         dev.write('SOUR:FREQ {}'.format(val['Hz']))
 
-
-
 __server__ = AgilentFunctionGenerator()
 
 if __name__ == '__main__':
     from labrad import util
     util.runServer(__server__)
-
-
-
--- a/agilent_33120a_function_gen.py
+++ b/agilent_33120a_function_gen.py
@@ -29,7 +29,7 @@ timeout = 20
 [shutdown]
 message = 987654321
 timeout = 20
-### END NODE INFO:b
+### END NODE INFO
 """
 
 from labrad import types as T, errors
--- a/agilent_33120a_function_gen.py
+++ b/agilent_33120a_function_gen.py
@@ -29,7 +29,7 @@ timeout = 20
 [shutdown]
 message = 987654321
 timeout = 20
-### END NODE INFO
+### END NODE INFO:b
 """
 
 from labrad import types as T, errors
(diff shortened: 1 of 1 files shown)


agilent_33120a_function_gen.py
```diff
@@ -38,25 +38,72 @@
...
 class AgilentFunctionGenerator(GPIBManagedServer):
     name = 'Agilent 33120a generator'
     deviceName = 'HEWLETT-PACKARD 33120A'
-    @setting(11,'Set DC', val='v[V]', returns='')
-    def set_dc_waveform(self, c,val=0*V):
-        if val< -5 or val > 5:
-            raise Exception('Signal Gnerator only puts out -5 to 5 volts in DC Voltage')
-        """Puts generator into DC mode with given voltage."""
+    
+    @setting(11, 'clear')
```
maffoo: Please remove trailing spaces.

agilent_33120a_function_gen.py
```diff
@@ -38,25 +38,72 @@
...
+        """
+
         allowed = ['50', 'INF']
         if setting not in allowed:
             raise Exception('allowed settings are: %s' % allowed)
         dev = self.selectedDevice(c)
-        dev.write('OUTP:LOAD %s' % setting)
+        dev.write('OUTP:LOAD {}'.format(val))
+
+    @setting(13,'Set DC voltage', val='v[V]', returns='')
```
maffoo: You changed the name here (was "Set DC") and the setting id. The id is generally not a problem because we call these by name, but changing the name could break people's code. Is there a reason for the change?

agilent_33120a_function_gen.py
```diff
@@ -38,25 +38,72 @@
...
+        """
+
         dev = self.selectedDevice(c)
-        dev.write(':APPL:DC DEF, DEF, %f' % float(val))
+        dev.write('*CLS')
       
-    @setting(12, 'set impedance', setting='s', returns='')      
-    def set_impedance(self, c, setting = '50'):
+    @setting(12, 'set impedance', val='s', returns='')      
+    def set_impedance(self, c, val = '50'):
```
maffoo: nit: no whitespace around `=` on parameter default values.

agilent_33120a_function_gen.py
```diff
@@ -38,25 +38,72 @@
...
+        """Sets the loading impedance
+        
+        args: 
+           val (str): The loading impedance
```
maffoo: Would be good to mention the allow values here, either '50' or 'INF'

agilent_33120a_function_gen.py
```diff
@@ -38,25 +38,72 @@
...
+
         allowed = ['50', 'INF']
         if setting not in allowed:
             raise Exception('allowed settings are: %s' % allowed)
         dev = self.selectedDevice(c)
-        dev.write('OUTP:LOAD %s' % setting)
+        dev.write('OUTP:LOAD {}'.format(val))
+
+    @setting(13,'Set DC voltage', val='v[V]', returns='')
+    def set_dc_voltage(self, c,val=0*V):
```
maffoo: nit: space after comma

agilent_33120a_function_gen.py
```diff
@@ -38,25 +38,72 @@
...
+        """Puts generator into DC mode with given voltage.
+        
+        args: 
+            val (value[V]): The DC Voltage to be set
+
+        """
+        if val< -5*U.V or val > 5*U.V:
```
maffoo: nit: whitespace

agilent_33120a_function_gen.py
```diff
@@ -38,25 +38,72 @@
...
+            raise Exception(
+                    'Signal Gnerator only puts out -5 to 5 volts in DC Voltage')
+        dev = self.selectedDevice(c)
+        dev.write('APPL:DC')
+        dev.write('VOLT:OFFS {}'.format(val['V']))
+    
+    @setting(14, 'set AC voltage', val='v[V]', returns='')      
+    def set_ac_voltage(self, c, val = '1*U.V'):
```
maffoo: nit: whitespace around `=`

agilent_33120a_function_gen.py
```diff
@@ -38,25 +38,72 @@
...
-        dev.write(':APPL:DC DEF, DEF, %f' % float(val))
+        dev.write('*CLS')
       
-    @setting(12, 'set impedance', setting='s', returns='')      
-    def set_impedance(self, c, setting = '50'):
+    @setting(12, 'set impedance', val='s', returns='')      
+    def set_impedance(self, c, val = '50'):
+        """Sets the loading impedance
+        
+        args:
```
maffoo: nit: "args" should be capitalized.

agilent_33120a_function_gen.py
```diff
@@ -38,25 +38,72 @@
...
-        dev.write('OUTP:LOAD %s' % setting)
+        dev.write('OUTP:LOAD {}'.format(val))
+
+    @setting(13,'Set DC voltage', val='v[V]', returns='')
+    def set_dc_voltage(self, c,val=0*V):
+        """Puts generator into DC mode with given voltage.
+        
+        args: 
+            val (value[V]): The DC Voltage to be set
+
```
maffoo: nit: remove blank line

agilent_33120a_function_gen.py
```diff
@@ -38,25 +38,72 @@
...
+        dev.write('VOLT:OFFS {}'.format(val['V']))
+    
+    @setting(14, 'set AC voltage', val='v[V]', returns='')      
+    def set_ac_voltage(self, c, val = '1*U.V'):
+        """Puts generator into AC mode with given peak to peak voltage.
+        
+        args: 
+            val (value[V]): The AC Voltage to be set
+        """
+
```
maffoo: nit: remove blank line

agilent_33120a_function_gen.py
```diff
@@ -38,25 +38,72 @@
...
+    def set_dc_voltage(self, c,val=0*V):
+        """Puts generator into DC mode with given voltage.
+        
+        args: 
+            val (value[V]): The DC Voltage to be set
+
+        """
+        if val< -5*U.V or val > 5*U.V:
+            raise Exception(
+                    'Signal Gnerator only puts out -5 to 5 volts in DC Voltage')
```
maffoo: nit: spelling.

agilent_33120a_function_gen.py
btchiaro: done

agilent_33120a_function_gen.py
btchiaro: done

agilent_33120a_function_gen.py
btchiaro: fixed

agilent_33120a_function_gen.py
btchiaro: got it.

maffoo:
I made some comments; mostly formatting nits. I know these are picky but it _really_ helps to be consistent everywhere (at least with new code...).

My only concern is with the setting name change; is changing the name going to affect many people? Also, with setting names more generally, I might suggest just leaving off the name for settings altogether in which case the python method name will be used for the setting too. I think it was a bad idea originally to get in the habit of making setting names with spaces and caps, only to then mangle these for the purpose of making them callable from python. So instead of 

```python
@setting(14, 'set AC voltage', val='v[V]', returns='')
def set_ac_voltage(self, c, val = '1*U.V'):
    ...
```

just do

```python
@setting(14, val='v[V]', returns='')
def set_ac_voltage(self, c, val = '1*U.V'):
    ...
```

Then the name will be "set_ac_voltage" whether referring to this setting by string name, or as a python method on the server object with the pylabrad client. We should have done this all along, but for new settings we can certainly adopt this style.

agilent_33120a_function_gen.py
btchiaro: I just wanted to be explicit and consistent with the method name.  Also when I set the AC parameters I need to specify either frequency or voltage, and I wanted the DC case to follow that form.

agilent_33120a_function_gen.py
btchiaro: got it

agilent_33120a_function_gen.py
btchiaro: Fixed

agilent_33120a_function_gen.py
btchiaro: ok

btchiaro:
Hey @maffoo I made the changes that you suggested.  I adopted the new style setting name convention everywhere, but I can change the set DC one back if you want.  I think I'm the only one using this server at the moment, so I don't know if the name change would cause problems.  I think its nice to be explicit about exactly what is being set and consistent with the AC case.

btchiaro:
Hey @maffoo were you ok with this given my comment above?  Or do you want me to change the "Set DC" back?  Thanks.

maffoo:
@btchiaro, sorry, I think I had a comment written and must have forgotten to actually submit it. Yes, LGTM.

PR #358, opened 2016-07-18
U/brooks/issue#356

BrooksCampbell:
This cleans up the Agilient Infiniium Scope server and adds a read_raw to the gpib server that does not .strip() the binary byte string.  This is necessary to fix an issue where data bytes are stripped if they correspond to a termination character.


gpib_server.py
```diff
@@ -41,6 +41,7 @@
 from twisted.internet.reactor import callLater
 from labrad.errors import DeviceNotSelectedError
 import labrad.units as units
+import numpy as np
```
maffoo: not needed

gpib_server.py
```diff
@@ -112,7 +113,7 @@ def refreshDevices(self):
                     elif addr.startswith('TCPIP'):
                         instName = addr
                     elif addr.startswith('USB'):
-                        instName = addr + '::INSTR'
+                        instName = addr
```
maffoo: this change seems unrelated.

gpib_server.py
```diff
@@ -194,6 +195,20 @@ def query(self, c, data):
         ans = instr.read_raw()
         return str(ans).strip()
 
+    @setting(7, n_bytes='w', returns='s')
```
maffoo: Should use the new bytes type "y" for the return type here, since this can be binary data. We'll have to think about modifying other settings in this server, but would be good to get this one right.

gpib_server.py
```diff
@@ -194,6 +195,20 @@ def query(self, c, data):
...
+    def read_raw(self, c, n_bytes=None):
+        """Read from the GPIB bus.  No .strip()
```
maffoo: Saying "No .strip()" will probably be confusing to clients looking at the labrad setting docs who don't have the source code in front of them and don't know the history of these changes. I would suggest saying something like:
```
"""Read raw bytes from the GPIB bus.

Termination characters, if any, will not be stripped.
If n_bytes is specified, reads only that many bytes.
Otherwise, reads until the device stops sending.
```

Then, we should modify the `read` setting docs to indicate that it _does_ strip termination characters.

gpib_server.py
```diff
@@ -194,6 +195,20 @@ def query(self, c, data):
...
+
+        If specified, reads only the given number of bytes.
+        Otherwise, reads until the device stops sending.
+        """
+        instr = self.getDevice(c)
+        if n_bytes is None:
+            ans = instr.read_raw()
+        else:
+            ans = instr.read_raw(n_bytes)
+        return str(ans)
```
maffoo: As per comment above, I'd suggest saying `bytes(ans)` here. In python 2 `bytes` is actually equal to `str`, but it's still a helpful hint IMO that this can be arbitrary bytes, not just a text string. This distinction will become more important in python3, of course.

agilent_infiniium_scope.py
```diff
@@ -38,29 +38,29 @@
...
 class AgilentDSO91304AServer(GPIBManagedServer):
     name = 'Agilent Infiniium Oscilloscope'
     deviceName = ['Agilent Technologies DSO91304A', 'KEYSIGHT TECHNOLOGIES DSO90804A']
-        
+
     @setting(11, returns=[])
     def reset(self, c):
         dev = self.selectedDevice(c)
-        yield dev.write('*RST')
+        yield dev.query('*RST;*OPC?')
```
maffoo: Is the `*OPC?` query here and below related to the binary parsing issue?

agilent_infiniium_scope.py
```diff
@@ -207,43 +207,48 @@ def averagemode(self, c, mode = None):
...
+        returnValue(navg_out)
 
     @setting(131, channel = 'i', level = 'v', returns = 'v{level}')
-    def trigger_at(self, c, channel, level = None):
-        """Get or set the trigger source and the trigger voltage for edge mode triggering.
-        Channel must be one of 0 (AUX), 1, 2, 3, 4.
+    def trigger_at(self, c, channel, level=None):
+        """Get or set the trigger source and the trigger voltage for edge mode
+         triggering.
+        Channel must be one of 0 (AUX), 1, 2, 3, 4, 5(LINE).
```
maffoo: Nit: docstring format says one line summary, then blank line, then rest. You can shorten the first line a bit, e.g. 
```python
"""Get or set the trigger source and voltage for edge triggering.

Channel must be one of 0 (AUX), 1, 2, 3, 4, or 5 (LINE).
"""

File: agilent_infiniium_scope.py [hunk 8]
```diff
@@ -207,43 +207,48 @@ def averagemode(self, c, mode = None):
...
         """
         dev = self.selectedDevice(c)
-        if channel==0:
+        if channel not in [0, 1, 2, 3, 4, 5]:
+            raise Exception('Select valid trigger channel')
```
maffoo: It's helpful to indicate the value that was given
```
raise ValueError('Invalid trigger channel: {}'.format(channel))
```
You can also include the allowed values, though they're in the setting docstring, so that should be sufficient.

agilent_infiniium_scope.py
```diff
@@ -207,43 +207,48 @@ def averagemode(self, c, mode = None):
...
+        if channel == 0:
             channel = 'AUX'
+        elif channel == 5:
+            channel = 'LINE'
         elif isinstance(channel, int):
-            #channel = 'CHAN%d' %channel
+            channel = 'CHAN{}'.format(channel)
             pass
```
maffoo: `pass` is no longer needed

agilent_infiniium_scope.py
```diff
@@ -207,43 +207,48 @@ def averagemode(self, c, mode = None):
...
+        yield dev.write('TRIG:EDGE:SOUR {}'.format(channel))
+        if channel != 'LINE':
+            yield dev.write('TRIG:LEV {}, {}'.format(channel, level))
+            #set trigger level
+            resp = yield dev.query('TRIG:LEV? {}'.format(channel))
         else:
-            yield dev.write('TRIG:EDGE:SOUR CHAN%s' %channel) #set channel up for edge triggering  
-            yield dev.write('TRIG:LEV CHAN%s,%f' %(channel,level)) #set trigger level
-            resp = yield dev.query('TRIG:LEV? CHAN%s' %channel)
+            resp=0.0
```
maffoo: nit: whitespace around `=`

agilent_infiniium_scope.py
```diff
@@ -366,18 +371,27 @@ def get_trace(self, c, channel, start=1, stop=10000):
         #Transfer waveform preamble
         preamble = yield dev.query('WAV:PRE?')
         #Transfer waveform data
-        binary = yield dev.query('WAV:DATA?')
+        p = dev._packet(). \
+            write('WAV:DATA?'). \
+            read_raw()
```
maffoo: styleguide says: no backslash line continuation (https://google.github.io/styleguide/pyguide.html#Line_length)

In this case, seems fine to just put on one line
```python
p = dev._packet().write('WAV:DATA?').read_raw()
```

agilent_infiniium_scope.py
```diff
@@ -459,17 +474,17 @@ def _parseBinaryData(data, wordLength):
     if wordLength == 1:
         lenHeader = int(data[1])
         dat = data[(2+lenHeader):]
-        dat = numpy.array(unpack(formatChar*(len(dat)/wordLength),dat))
+        dat = np.array(unpack(formatChar*(len(dat)/wordLength),dat))
```
maffoo: nit: whitespace around `*` and after comma

agilent_infiniium_scope.py
```diff
@@ -366,18 +371,27 @@ def get_trace(self, c, channel, start=1, stop=10000):
...
         preamble = yield dev.query('WAV:PRE?')
         #Transfer waveform data
-        binary = yield dev.query('WAV:DATA?')
+        p = dev._packet(). \
+            write('WAV:DATA?'). \
+            read_raw()
+
+        result = yield p.send()
+        binary = result['read_raw']
+        binary = binary[:-1]
```
maffoo: Why are you dropping the last byte?

agilent_infiniium_scope.py
```diff
@@ -366,18 +371,27 @@ def get_trace(self, c, channel, start=1, stop=10000):
...
         numPoints = int(preambleDict['numPoints'])
-        time = numpy.linspace(float(preambleDict['xFirst']),
-                              (numPoints-1) * float(preambleDict['xStep'])+float(preambleDict['xFirst']),
-                              numPoints)
+
+        print 'Expected {} points, got {} point'.format(numPoints, len(traceVolts))
+
+        time = np.linspace(float(preambleDict['xFirst']),
+                           (numPoints-1) * float(preambleDict['xStep'])+float(preambleDict['xFirst']),
+                           numPoints)
```
maffoo: I would pull out the preamble params first:
```python
first, step = float(preambleDict['xFirst']), float(preambleDict['xStep'])
time = np.linspace(first, first + (numPoints-1) * step, numPoints)
```

agilent_infiniium_scope.py
```diff
@@ -366,18 +371,27 @@ def get_trace(self, c, channel, start=1, stop=10000):
...
+
+        result = yield p.send()
+        binary = result['read_raw']
+        binary = binary[:-1]
         #Parse waveform preamble
         preambleDict = _parsePreamble(preamble)
         print preambleDict
         #Parse binary
-        trace = _parseBinaryData(binary, wordLength = wordLength) * 1.0e3
+        trace = _parseBinaryData(binary, wordLength = wordLength)
```
maffoo: nit: no space around `=` for keyword args

maffoo:
@BrooksCampbell, thanks for doing this! It would be good to break this up into several changes; in particular, I think we should do the gpib server changes first, and we need to bump the version number since this adds a new setting. Probably want to bump the minor version, so 1.4.0-no-refresh.

As for the scope server, there seem to be several different changes going on: whitespace and formatting stuff, then binary trace stuff, then triggering. It's always helpful to keep changes small and logically related, expecially when you want to go back through the commit history and see what happened when. We'll also want to bump the scope server version, though there are no interface changes there so can just change the patch number, e.g. 0.2.1.

Let me know if you need help with git to split this up.

agilent_infiniium_scope.py
BrooksCampbell: It is not related, but an improvement to the server after talking to Keysight support.  The *OPC? query confirms that the scope has finished parsing commands.  It does not mean that the scope has actually finished executing the previous command.  Keysight support recommended adding in some *OPC? queries in order to give the scope time to execute some of the longer commands.

agilent_infiniium_scope.py
BrooksCampbell: Probably not the correct way to handle this, but I believe that there is a single termination character at the end of the byte string so I drop it manually.

agilent_infiniium_scope.py
```diff
@@ -227,28 +227,30 @@ def numavg(self, c, navg = None):
...
+
+        Channel must be one of 0 (AUX), 1, 2, 3, 4, or 5 (LINE).
+        If trigger source is 5 (LINE) the level will be ignored.
         """
         dev = self.selectedDevice(c)
         if channel not in [0, 1, 2, 3, 4, 5]:
-            raise Exception('Select valid trigger channel')
+            raise ValueError('Invalid trigger channel: {}'
+                             'Valid channels are [0, 1, 2, 3, 4, 5]'
+                             '0 for AUX, 5 for LINE'.format(channel))
```
maffoo: When you have neighboring string literals like this, they get concatenated without additional spaces or newlines, so you'll want to add some punctuation and spaces:
```python
'Invalid trigger channel: {}. '
'Valid channels are [0, 1, 2, 3, 4, 5] '
'(0 for AUX, 5 for LINE).'.format(channel)
```

agilent_infiniium_scope.py
```diff
@@ -227,28 +227,30 @@ def numavg(self, c, navg = None):
...
-            pass
 
         yield dev.write('TRIG:EDGE:SOUR {}'.format(channel))
         if channel != 'LINE':
             yield dev.write('TRIG:LEV {}, {}'.format(channel, level))
             #set trigger level
             resp = yield dev.query('TRIG:LEV? {}'.format(channel))
         else:
-            resp=0.0
+            resp=  0.0
```
maffoo: check whitespace here.

agilent_infiniium_scope.py
```diff
@@ -371,27 +373,26 @@ def get_trace(self, c, channel, start=1, stop=10000):
...
-        print preambleDict
+
         #Parse binary
-        trace = _parseBinaryData(binary, wordLength = wordLength)
+        trace = _parseBinaryData(binary, wordLength=wordLength)
+
         #Convert from binary to volts
-        traceVolts = ((trace*float(preambleDict['yStep'])+float(preambleDict['yOrigin'])))
+        step, origin = (float(preambleDict['yStep']),
+                        float(preambleDict['yOrigin']))
```
maffoo: If this doesn't fit on one line, I'd suggest just making two separate assignment statements:
```python
step = float(preambleDict['yStep'])
origin = float(preambleDict['yOrigin'])
```

agilent_infiniium_scope.py
```diff
@@ -371,27 +373,26 @@ def get_trace(self, c, channel, start=1, stop=10000):
...
+        traceVolts = (trace * step) + origin
         numPoints = int(preambleDict['numPoints'])
 
-        print 'Expected {} points, got {} point'.format(numPoints, len(traceVolts))
-
-        time = np.linspace(float(preambleDict['xFirst']),
-                           (numPoints-1) * float(preambleDict['xStep'])+float(preambleDict['xFirst']),
-                           numPoints)
+        first, step = (float(preambleDict['xFirst']),
+                       float(preambleDict['xStep']))
```
maffoo: same comment as above

maffoo:
A couple of minor comments. Also please break this up into separate logical changes and bump server versions.

BrooksCampbell:
Okay, as my last act before vacation, I want to get this pull request done.

1.  How do I bump server versions?
2.  When you say to break this up into separate logical changes, do you just want separate commits for each of the logical changes, or do you want me to make separate pull requests for each of the logical changes?

Let me know if there is a better way to do this, but since I believe that currently I have single commits that include multiple logical changes, I'm going to have to do some manual editing, right?  I think I know how to go and pull out the individual commits and make separate pull requests for each (for the cases where my logical changes are properly separated into individual commits).

jwenner:
To change the server version, change [this line](https://github.com/martinisgroup/servers/blob/4cd2d3ad10df321dea6cb7d7411f410e49d08a99/agilent_infiniium_scope.py#L20).

BrooksCampbell:
I moved these changes in to three different pull requests so I am closing this pull request.

Issue #288, opened 2015-12-10
Accepted parameters of the new qubit server doesn't compatible with the old version


SchineCompton:
Hi,

I successfully build the latest scala version of QubitServer and struggled for a few days for a flatten error. Finally I found that the function `new_sram_block` accepts `s,w,(ss)` and  the old version of qubitserver accepts only `s,w`, so our code is no longer compatible with the new QubitServer . What does the added parameter mean ? Do you have any document about it ?  Or may I have some demo code to understand how to control the new QubitServer ?

There is another question that we are using the old version of FPGA board, DAC build 8 and ADC build 1. Is the qubit server compatible with our board ?

Here is our old version that only accepts "s,w".

``` python
In [58]: cxn.qubit_sequencer.new_sram_block?
Type:           SettingWrapper
Base Class:     <class 'labrad.client.SettingWrapper'>
String Form:
LabRAD Setting: "Qubit Sequencer" >> "New SRAM Block" (ID=400)

            Create a new SRAM block with the  <...> a will be padded at the begi
nning after deconvolution.

            Accepts:
            s, w: s, w

            Returns:
```

Thanks

DanielSank:
+1

The documentation is vague.

Coxious:
We are currently using the old pyle that doesn't compatible with latest qubit server. Can we have a look at the pyle project you currently using ?

DanielSank:
> We are currently using the old pyle that doesn't compatible with latest qubit server. Can we have a look at the pyle project you currently using?

That code is not public. Let's focus on improving the documentation for the qubit sequencer so that you can update your python code appropriately.

Coxious:
Okay, thanks anyway.

Coxious:
There is another question, we are using the old version of FPGA board, DAC build 8 and ADC build 1. Is the scala QubitServer compatible with the old board ?

DanielSank:
I think so. @pomalley?

Coxious:
Oh that is great. Then maybe I can still change my code to adapt the QubitServer. Thanks !

pomalley:
It is supposed to be compatible, yes, but I don't think it's been tested since we tried it out once about nine or ten months ago.

maffoo:
[...]

PR #361, opened 2016-07-27
Remove old versions of direct ethernet

ejeffrey:
These are out of date.


jwenner:
I strongly disagree with this PR. This is the last publicly released version of the Direct Ethernet server. If we remove this, how are groups outside of ours supposed to talk with their DAC/ADC boards?

ejeffrey:
If the up to date code is not here, they need to be getting it from us directly.  Or we should move the code from pyle back here.

It will still, of course, be available in the history, but by publishing obsolete code doesn't help anyone.

maffoo:
@jwenner, also, the binaries that we've built are published on bintray (https://bintray.com/labrad/generic/direct-ethernet) and at least the last release of the direct ethernet server was tagged as [direct-ethernet-v1.2.3](https://github.com/martinisgroup/servers/tree/direct-ethernet-v1.2.3) so it is here in the repo and always possible to go back. So nothing is going away for people who need it, this would just be cleaning up the tip of tree to only contain stuff that is being actively developed.

I do agree that we need to discuss the best way to deal with things here, however, because this repository is kind of a hodgepodge, and different people are probably using the code in different ways.

jwenner:
Please see rationale for having the updated code in a private repo here: martinisgroup/pyle#1073.

And as an example of another such group using the qubit sequencer code, Haohua's group (see #288), and Robert's group uses the FPGA server and so direct ethernet server.

Now, what I would support is putting a note in the readmes saying that these servers are no longer being updated and to contact us for the latest versions.

ejeffrey:
OK, I am going to leave the fpga server and calibration server for now (but remember, the calibrations server is broken!).

The direct ethernet server and qubit server I am going to remove from master, but add a readme with instructions pointing to both the bintray release and the release tag for the source repo.

ejeffrey:
OK, notes added with both bintray and github links.

I really don't want someone checking this out and thinking it is the latest repo, and by referring to a tag, that will prevent anyone from making modifications and trying to submit a PR which we can't accept.

Readme_Direct_Ethernet.txt
```diff
@@ -0,0 +1,11 @@
+The direct ethernet is no longer being actively developed here. To
+avoid confusion it has been removed from master.
+
+The latest binary release (1.2.3) is available via bintray:
+
+https://bintray.com/labrad/generic/direct-ethernet
+
+The source code is available in a release tag of the git repository:
+
+https://github.com/martinisgroup/servers/tree/direct-ethernet-v1.2.3
```
maffoo: I would suggest putting all these notes in the main `README.md` so this is shows up for people browsing the repo in the web interface.

Readme_Direct_Ethernet.txt
```diff
@@ -0,0 +1,11 @@
+The direct ethernet is no longer being actively developed here. To
+avoid confusion it has been removed from master.
+
+The latest binary release (1.2.3) is available via bintray:
+
+https://bintray.com/labrad/generic/direct-ethernet
+
+The source code is available in a release tag of the git repository:
```
maffoo: Here, can just say that the latest release is at tag `direct-ethernet-v1.2.3`. Can give the github link as well, but people using git command line tools just need to know the tag name.

Readme_Direct_Ethernet.txt
ejeffrey: I will mention it in README.md, but someone who checks out the repo and is looking for the files may not see it there.

DirectEthernet/README.txt
```diff
@@ -0,0 +1,12 @@
+The direct ethernet is no longer being actively developed here. To
```
maffoo: Maybe makes these `.md` files. Github renders them nicely (with links, etc) and they're still perfectly readable as plain text.

ejeffrey:
OK, converted to .md files

maffoo:
LGTM

PR #362, opened 2016-07-27
Remove obsolete qubit sequencer

ejeffrey:


maffoo:
LGTM

jwenner:
Are there other groups using the public qubit sequencer server? If so, then we shouldn't remove it from the repo, as the more current version is private.

ejeffrey:
My opinion is that by moving this code into pyle, we are saying that anyone external who uses this should be getting the code via special arrangement.

PR #363, opened 2016-07-27
Remove obsolete GHz dac calibration / server

ejeffrey:
Need to make sure people aren't running this code from working copies before merging, and send out an announcement.


jwenner:
Note that, unlike what @maffoo said in #361, there is no publicly-accessible copy of this server outside of this repo.

jwenner:
At a minimum, if we want to clean up this repo, we should move these servers (qubit sequencer, FPGA, direct ethernet, DAC deconvolution) over to their own repo.

PR #365, opened 2016-07-29
Fix gpib addressing to remove redundent ::INSTR.

BrooksCampbell:
Breaking up logical chages from #358.

This branch removes a redundant ::INSTR we were adding to USB devices in the gpib server.

I don't know that this is always the correct thing to do, but it definitely is for the two USB scopes I've used:  Agilent DSA90804A scope and DSO-X2014A


gpib_server.py
```diff
@@ -112,7 +112,7 @@ def refreshDevices(self):
                     elif addr.startswith('TCPIP'):
                         instName = addr
                     elif addr.startswith('USB'):
-                        instName = addr + '::INSTR'
+                        instName = addr
```
maffoo: Is the problem that `addr` already includes `::INSTR`, so adding this again is not needed? Or is the problem that the instrument identifier doesn't need the suffix `::INSTR` at all?

If we do this, then the three branches can all be collapsed into check:
```python
KNOWN_DEVICE_TYPES = ('GPIB', 'TCPIP', 'USB')
...
if not addr.startswith(KNOWN_DEVICE_TYPES):
    continue
instr = rm.get_instrument(addr)
...
```

maffoo: I learned about passing a tuple of strings to `startswith` in #205.

BrooksCampbell: I jut double checked and the the USB scope's address already ends in '::INSTR'.

Is "KNOWN_DEVICE_TYPES" a global?  If so, I think I made these changes correctly in the next commit.

maffoo: Yeah, I was imagining that tuple as being a module-level global, just like you did it.

gpib_server.py
```diff
@@ -62,6 +62,7 @@
 ### END NODE INFO
 """
 
+KNOWN_DEVICE_TYPES = ('GPIB', 'TCPIP', 'USB')
 
 class GPIBBusServer(LabradServer):
```
maffoo: nit: two blank lines around top level definitions (https://google.github.io/styleguide/pyguide.html#Blank_Lines)

maffoo:
One formatting nit, then LGTM.

PR #367, opened 2016-07-29
U/brooks/infiniium scope server cleanup

BrooksCampbell:
Update infiniium scope server to use read_raw to pull binary data.

Adds two *OPC? queries to the *RST and *CLS writes (Keysight suggested this as a best practice).

Also fixed a lot of white space issues and camel case variable names.


agilent_infiniium_scope.py
```diff
@@ -31,39 +31,39 @@
...
-        
+    deviceName = ['Agilent Technologies DSO91304A',
+                  'KEYSIGHT TECHNOLOGIES DSO90804A']
+
     @setting(11, returns=[])
     def reset(self, c):
         dev = self.selectedDevice(c)
-        yield dev.write('*RST')
+        yield dev.query('*RST;*OPC?')
         # TODO wait for reset to complete
```
maffoo: With the addition of `*OPC?`, does this now wait until the rest completes? If so, you can remove the TODO.

agilent_infiniium_scope.py
```diff
@@ -72,319 +72,350 @@ def channel_infoNONEXISTANT(self, c, channel):
...
+        resp = yield dev.query('CH{}?'.format(channel))
+        (bwLimit, coupling, deskew, offset, invert, position, scale,
+         termination, probeCal, probeAtten, resistance, unit, textID,
+         textSN, extAtten, extUnits, textLabel, xPos, yPos) = resp.split(';')
+
+        # Convert strings to numerical data when appropriate
+        probeAtten = T.Value(float(probeAtten), '')
+        termination = T.Value(float(termination), '')
+        scale = T.Value(float(scale), '')
+        position = T.Value(float(position), '')
```
maffoo: Since these are dimensionless, can just use floats: `probeAtten = float(probeAtten)`, etc.

agilent_infiniium_scope.py
```diff
@@ -72,319 +72,350 @@ def channel_infoNONEXISTANT(self, c, channel):
...
         coupling = coupling
-        bwLimit = T.Value(float(bwLimit),'')
+        bwLimit = T.Value(float(bwLimit), '')
         invert = invert
-        unit = unit[1:-1] #Get's rid of an extra set of quotation marks
+        unit = unit[1:-1]  # Get's rid of an extra set of quotation marks
```
maffoo: nit: no apostrophe

agilent_infiniium_scope.py
```diff
@@ -72,319 +72,350 @@ def channel_infoNONEXISTANT(self, c, channel):
...
-
-        #Convert strings to numerical data when appropriate
-        probeAtten = T.Value(float(probeAtten),'')
-        termination = T.Value(float(termination),'')
-        scale = T.Value(float(scale),'')
-        position = T.Value(float(position),'')
+        resp = yield dev.query('CH{}?'.format(channel))
+        (bwLimit, coupling, deskew, offset, invert, position, scale,
+         termination, probeCal, probeAtten, resistance, unit, textID,
+         textSN, extAtten, extUnits, textLabel, xPos, yPos) = resp.split(';')
```
maffoo: Ugh, that's gross. I guess this formatting is all a bit academic given that the function throws an exception on the first line, though...

agilent_infiniium_scope.py
```diff
@@ -72,319 +72,350 @@ def channel_infoNONEXISTANT(self, c, channel):
...
-        #'1.0E1;1.0E1;2.5E1;0.0E0;DC;OFF;OFF;"V"'
-        #These strings represent respectively,
-        #probeAttenuation;termination;vertScale;vertPosition;coupling;bandwidthLimit;invert;vertUnit
+        """
+        NOTES
+        The scope's response to 'CH<x>?' is a string of format
+        '1.0E1;1.0E1;2.5E1;0.0E0;DC;OFF;OFF;"V"'
+        These strings represent respectively,
+        probeAttenuation;termination;vertScale;vertPosition;coupling;
+        bandwidthLimit;invert;vertUnit
```
maffoo: This comment does not agree with the unpacking that happens a few lines below. Why is that? Also, apart from docstrings, I'd suggest leaving `#` for comments, rather than just putting string literals in other parts of the code.

agilent_infiniium_scope.py
BrooksCampbell: Not technically... *OPC?  just waits for the scope to parse the commands... Keysight's applicaiton engineer said that it is "usually good enough" and didn't suggest any better method.

There may or may not be some status bit we could check, but I don't think I was having an issue before adding the *OPC? query and adding it should just make things better.

agilent_infiniium_scope.py
```diff
@@ -72,319 +72,350 @@ def channel_infoNONEXISTANT(self, c, channel):
...
         dev = self.selectedDevice(c)
-        if scale is None:
-            resp = yield dev.query('CHAN%d:SCAL?' %channel)
-        else:
-            scale = format(scale,'E')
-            yield dev.write(('CHAN%d:SCAL '+scale) %channel)
-            resp = yield dev.query('CHAN%d:SCAL?' %channel)
+        if scale is not None:
+            scale = format(scale, 'E')
+            yield dev.write('CHAN{}:SCAL {}'.format(channel, scale))
```
maffoo: You can put the format code right in the formatting string:
```python
if scale is not None:
    yield dev.write('CHAN{}:SCAL {:E}'.format(channel, scale))
```

agilent_infiniium_scope.py
```diff
@@ -72,319 +72,350 @@ def channel_infoNONEXISTANT(self, c, channel):
...
-        chString = 'CH%d:' %channel
+        ch_string = 'CH{}:'.format(channel)
         if factor is None:
-            resp = yield dev.query(chString+'PRO?')
+            resp = yield dev.query('{}PRO?'.format(ch_string))
         elif factor in probeFactors:
-            yield dev.write(chString+'PRO %d' %factor)
-            resp = yield dev.query(chString+'PRO?')
+            yield dev.write('{}PRO {}'.format(ch_string, factor))
+            resp = yield dev.query('{}PRO?'.format(chString))
```
maffoo: `ch_string`

agilent_infiniium_scope.py
```diff
@@ -72,319 +72,350 @@ def channel_infoNONEXISTANT(self, c, channel):
...
         raise Exception('Not yet implemented')
-        probeFactors = [1,10,20,50,100,500,1000]
+        probe_factors = [1, 10, 20, 50, 100, 500, 1000]
         dev = self.selectedDevice(c)
-        chString = 'CH%d:' %channel
+        ch_string = 'CH{}:'.format(channel)
         if factor is None:
-            resp = yield dev.query(chString+'PRO?')
+            resp = yield dev.query('{}PRO?'.format(ch_string))
         elif factor in probeFactors:
```
maffoo: `probe_factors`

agilent_infiniium_scope.py
BrooksCampbell: Fixed.

agilent_infiniium_scope.py
```diff
@@ -72,319 +72,350 @@ def channel_infoNONEXISTANT(self, c, channel):
...
+    @setting(118, mode=['s', 'i', 'b'], returns=['s'])
+    def averagemode(self, c, mode=None):
         """Get or set acquisition mode
         """
         dev = self.selectedDevice(c)
-        if mode is None:
-            resp = yield dev.query('ACQ:AVER?')
-        else:
+        if mode is not None:
             if mode is not None:
```
maffoo: redundant check on `mode` here

agilent_infiniium_scope.py
BrooksCampbell: cool, and fixed.

agilent_infiniium_scope.py
```diff
@@ -72,319 +72,350 @@ def channel_infoNONEXISTANT(self, c, channel):
...
-            pass
-        if channel not in [0,1,2,3,4]:
-            raise Exception('Select valid trigger channel')
-        if level is None:
-            resp = yield dev.query('TRIG:LEV? CHAN%s' %channel)
+            channel = 'CHAN{}'.format(channel)
+        yield dev.write('TRIG:EDGE:SOUR {}'.format(channel))
+        if channel != 'LINE':
+            yield dev.write('TRIG:LEV {}, {}'.format(channel, level))
+            # set trigger level
```
maffoo: nit: put this comment _before_ the previous line instead

agilent_infiniium_scope.py
BrooksCampbell: Whoops, this was the first one I did before I started find/replacing them...

agilent_infiniium_scope.py
```diff
@@ -72,319 +72,350 @@ def channel_infoNONEXISTANT(self, c, channel):
...
+    @setting(131, channel='i', level='v', returns='v{level}')
+    def trigger_at(self, c, channel, level=None):
+        """Get or set the trigger source and voltage for edge mode triggering.
 
-    @setting(131, channel = 'i', level = 'v', returns = 'v{level}')
-    def trigger_at(self, c, channel, level = None):
-        """Get or set the trigger source and the trigger voltage for edge mode triggering.
-        Channel must be one of 0 (AUX), 1, 2, 3, 4.
+        Channel must be one of 0 (AUX), 1, 2, 3, 4, or 5 (LINE).
+        If trigger source is 5 (LINE) the level will be ignored
```
maffoo: Can't you still set a trigger level when triggering off the line? Sorry, clearly it's been too long since I fiddled with a scope...

agilent_infiniium_scope.py
BrooksCampbell: What is the correct way to do it?

agilent_infiniium_scope.py
BrooksCampbell: I've fixed the # for the comments, but I do not know about the unpacking...

In fact, I dont think that the CH{}? command even exists for the Infiniium series of scopes.  I'm going to guess this was copied from another scope server and never implemented.  For these scopes this would be a series of individual commands.  Probably we should kill this function and we could add in different functions for each of these parameters, or just leave it out.

agilent_infiniium_scope.py
maffoo: I don't think this is incorrect, just saying it's hard to read.

agilent_infiniium_scope.py
BrooksCampbell: For these scopes at least.  I probably realized that by trial and error, but here is it confirmed from the manual:

http://www.keysight.com/upload/cmc_upload/All/9000_series_prog_ref.pdf#page=914

agilent_infiniium_scope.py
maffoo: Yeah, given that this throws a not implemented exception, and has the scary-looking name `channel_infoNONEXISTENT`, let's just get rid of this entire function. Same for the other `*NONEXISTENT` functions below.

agilent_infiniium_scope.py
```diff
@@ -72,319 +72,350 @@ def channel_infoNONEXISTANT(self, c, channel):
...
+        yield dev.write('TRIG:EDGE:SOUR {}'.format(channel))
+        if channel != 'LINE':
+            yield dev.write('TRIG:LEV {}, {}'.format(channel, level))
+            # set trigger level
+            resp = yield dev.query('TRIG:LEV? {}'.format(channel))
         else:
-            yield dev.write('TRIG:EDGE:SOUR CHAN%s' %channel) #set channel up for edge triggering  
-            yield dev.write('TRIG:LEV CHAN%s,%f' %(channel,level)) #set trigger level
-            resp = yield dev.query('TRIG:LEV? CHAN%s' %channel)
+            resp = 0.0
```
maffoo: I would change this condition around to put the special case of line triggering near the check:
```python
if channel == 'LINE':
    # Cannot set trigger level when triggering off line input.
    # See http://www.keysight.com/upload/cmc_upload/All/9000_series_prog_ref.pdf#page=914
    level = 0.0
else:
    # Set trigger level.
    yield dev.write('TRIG:LEV {}, {}'.format(channel, level))
    resp = yield dev.query('TRIG:LEV? {}'.format(channel))
    level = float(resp)
returnValue(level)
```

agilent_infiniium_scope.py
```diff
@@ -31,360 +31,389 @@
...
         """
         dev = self.selectedDevice(c)
         if slope is None:
             resp = yield dev.query('TRIG:MODE?')
         else:
             slope = slope.upper()
-            if slope not in ['COMM','DEL','EDGE','GLIT','PATT','PWID','RUNT','SEQ',',SHOL','STAT','TIM','TRAN','TV',',WIND','SBUS1','SBUS2','SBUS3','SBUS4']:
+            if slope not in ['COMM', 'DEL', 'EDGE', 'GLIT', 'PATT', 'PWID',
+                             'RUNT', 'SEQ', ',SHOL', 'STAT', 'TIM', 'TRAN',
+                             'TV', ',WIND', 'SBUS1', 'SBUS2', 'SBUS3', 'SBUS4']:
```
maffoo: Should be 'WIND', not ',WIND'. Also incorrect in docstring.

agilent_infiniium_scope.py
```diff
@@ -31,360 +31,389 @@
...
         """Get or set the horizontal scale
         """
         dev = self.selectedDevice(c)
         if scale is None:
             resp = yield dev.query('TIM:SCAL?')
         else:
-            scale = format(scale,'E')
-            yield dev.write('TIM:SCAL '+scale)
+            scale = format(scale, 'E')
+            yield dev.write('TIM:SCAL {}'.format(scale))
```
maffoo: Here again, just put formatting code in format string: `yield dev.write('TIM:SCAL {:E}'.format(scale))`

agilent_infiniium_scope.py
```diff
@@ -31,360 +31,389 @@
...
             resp = yield dev.query('TIM:SCAL?')
```
maffoo: This query happens on both branches, so can move it outside:
```python
if scale is not None:
    yield dev.write(...)
resp = yield dev.write('TIM:SCAL?')
returnValue(float(resp))
```

agilent_infiniium_scope.py
```diff
@@ -31,360 +31,389 @@
...
         returnValue(scale)
     
-    #Data acquisition settings
-    @setting(201, channel = 'i', start = 'i', stop = 'i', returns='*v[ns] {time axis} *v[mV] {scope trace}')
+    # Data acquisition settings
+    @setting(201, channel='i', start='i', stop='i',
+             returns='*v[ns] {time axis} *v[mV] {scope trace}')
     def get_trace(self, c, channel, start=1, stop=10000):
         """Get a trace from the scope.
         OUTPUT - (array voltage in volts, array time in seconds)
```
maffoo: nit: blank line after first line of docstring

agilent_infiniium_scope.py
BrooksCampbell: fixed.

agilent_infiniium_scope.py
```diff
@@ -31,360 +31,389 @@
...
         """
-        #raise Exception('Doesnt work yet. Please fix lines defining traceVolts and time.')
-##        DATA ENCODINGS
-##        RIB - signed, MSB first
-##        RPB - unsigned, MSB first
-##        SRI - signed, LSB first
-##        SRP - unsigned, LSB first
-        wordLength = 2 #Hardcoding to set data transer word length to 2 bytes
+        # raise Exception('Doesnt work yet. Please fix lines defining
+        # trace_volts and time.')
```
maffoo: remove commented-out exception

agilent_infiniium_scope.py
BrooksCampbell: Makes sense.   I checked the manual and these commands are not even really supported by these scopes--i.e. AC/DC coupling is only available through probes.

I have removed all the non-implemented functions.

agilent_infiniium_scope.py
BrooksCampbell: fixed

agilent_infiniium_scope.py
BrooksCampbell: fixed

agilent_infiniium_scope.py
```diff
@@ -31,360 +31,389 @@
...
-        time = numpy.linspace(float(preambleDict['xFirst']),
-                              (numPoints-1) * float(preambleDict['xStep'])+float(preambleDict['xFirst']),
-                              numPoints)
-
-        returnValue((time*U.ns*1e9, traceVolts*U.V))
+        # Transfer waveform data
+        p = dev._packet().write('WAV:DATA?').read_raw()
+        result = yield p.send()
+        binary = result['read_raw']
+        binary = binary[:-1]
```
maffoo: I would combine these, and add a comment about dropping the termination character:
```python
result = yield dev._packet().write('WAV:DATA?').read_raw()
binary = result['read_raw'][:-1]  # drop termination character
```

agilent_infiniium_scope.py
BrooksCampbell: Gotcha.

agilent_infiniium_scope.py
BrooksCampbell: I had noticed that for ',SHOL' too, but was lazy and didn't check the manual to make sure it was actually a type ;).  But now I did, and it is a typo.

Fixed now.

agilent_infiniium_scope.py
```diff
@@ -31,360 +31,389 @@
...
+        # Convert from binary to volts
+
+        y_step = float(preample_dict['yStep'])
+        origin = float(preample_dict['yOrigin'])
+        trace_volts = (trace * y_step) + origin
+
+        num_points = int(preample_dict['numPoints'])
+        x_step = float(preample_dict['xStep'])
+        first = float(preample_dict['xFirst'])
+        time = numpy.linspace(first, first + (num_points-1) * x_step, num_points)
```
maffoo: I would call this `time_s` since it is in units of seconds. Then:
```python
time_ns = time_s * 1e9
returnValue((time_ns*U.ns, trace_volts*U.V))
```
(time_s*U.s).inUnitsOf

agilent_infiniium_scope.py
```diff
@@ -409,67 +438,69 @@ def _parsePreamble(preamble):
                     ('maxBW',False),
                     ('minBW',False)]
     '''
```
maffoo: Remove this "commented" code

agilent_infiniium_scope.py
BrooksCampbell: Fixed.  I also searched for other instances but didn't find any.

agilent_infiniium_scope.py
```diff
@@ -409,67 +438,69 @@ def _parsePreamble(preamble):
...
-                ('model',True),
-                ('acquisitionMode',True),
-                ('percentTimeBucketsComplete',True),
-                ('xUnits',True),
-                ('yUnits',True),
-                ('maxBW',True),
-                ('minBW',True)]
-    preambleDict = {}
-    for key,val in zip(preambleKeys,preambleVals):
+    preamble_keys = [('byteFormat', True),
```
maffoo: This can be defined at the module level, rather than inside this function which rebuilds the list every time it is called. Also, I would prefer the following indentation:
```python
PREAMBLE_KEYS = [
    ('byteFormat', True),
    ...
]
```

agilent_infiniium_scope.py
```diff
@@ -409,67 +438,69 @@ def _parsePreamble(preamble):
...
+                ('time', True),
+                ('model', True),
+                ('acquisitionMode', True),
+                ('percentTimeBucketsComplete', True),
+                ('xUnits', True),
+                ('yUnits', True),
+                ('maxBW', True),
+                ('minBW', True)]
+    preample_dict = {}
+    for key, val in zip(preamble_keys, preamble_vals):
```
maffoo:
```python
for (key, included), val in zip(PREAMBLE_KEYS, preamble_vals):
    if included:
        preamble_dict[key] = val
```

agilent_infiniium_scope.py
```diff
@@ -409,67 +438,69 @@ def _parsePreamble(preamble):
...
+        elif num == '2':
             return 'ns'
         else:
             raise Exception('Units not time or voltage')
-    preambleDict['xUnit'] = unitType(preambleDict['xUnits'])
-    preambleDict['yUnit'] = unitType(preambleDict['yUnits'])
-    return (preambleDict)
+    preample_dict['xUnit'] = unit_type(preample_dict['xUnits'])
+    preample_dict['yUnit'] = unit_type(preample_dict['yUnits'])
+    return (preample_dict)
```
maffoo: parens not needed

agilent_infiniium_scope.py
BrooksCampbell: Noted.  I changed it for horiz_scale,  numavg, and position,  too. 

Is there a good way to do a similar simplification with this?

```
    @setting(150, side='s', returns=['s'])
    def horiz_refpoint(self, c, side=None):
        """Get or set the reference point for the horizontal position.

        Must be 'LEFT', 'CENT'er, or 'RIGH't.
        """
        dev = self.selectedDevice(c)
        if side is not None:
            resp = yield dev.query('TIM:REF?')
        else:
            side = side.upper()
            if side not in ['LEFT', 'CENT', 'RIGH']:
                raise Exception('Mode must be "LEFT", "CENT", or "RIGH".')
            else:
                yield dev.write('TIM:REF {}'.format(side))
                resp = yield dev.query('TIM:REF?')
        returnValue(resp)
```

agilent_infiniium_scope.py
BrooksCampbell: fixed.

agilent_infiniium_scope.py
```diff
@@ -409,67 +438,69 @@ def _parsePreamble(preamble):
...
-        dat = data[(2+lenHeader):]
-        dat = dat[-calcsize('>' + formatChar*(len(dat)/wordLength)):]
-        dat = numpy.array(unpack('>' + formatChar*(len(dat)/wordLength),dat))
-    elif wordLength == 4:
-        lenHeader = int(data[1])
-        dat = data[(2+lenHeader):]
-        dat = dat[-calcsize('>' + formatChar*(len(dat)/wordLength)):]
-        dat = numpy.array(unpack('>' + formatChar*(len(dat)/wordLength),dat))      
+    format_chars = {'1': 'b', '2': 'h', '4': 'f'}
+    format_char = format_chars[str(word_length)]
```
maffoo:
```python
format_chars = {1: 'b', 2: 'h', 4: 'f'}
format_char = format_chars[word_length]
```

agilent_infiniium_scope.py
BrooksCampbell: Does this send the packet still?

maffoo: Oh, sorry, you're right, would have to add a `.send()` on the end. Probably best to leave the packet construction and sending on separate lines, as you have it.

agilent_infiniium_scope.py
BrooksCampbell: done.

agilent_infiniium_scope.py
BrooksCampbell: done

agilent_infiniium_scope.py
maffoo: Can do exactly the same thing, because the code bails out in the branch with an exception. It's generally good to use this to not introduce extra levels of nesting, that is:
```python
# excess nesting
if condition:
    raise Exception()
else:
    do_something()

# do this instead
if condition:
    raise Exception()
do_something()
```
So for the code you quoted:
```python
    @setting(150, side='s', returns=['s'])
    def horiz_refpoint(self, c, side=None):
        """Get or set the reference point for the horizontal position.

        Must be 'LEFT', 'CENT'er, or 'RIGH't.
        """
        dev = self.selectedDevice(c)
        if side is None:
            side = side.upper()
            if side not in ['LEFT', 'CENT', 'RIGH']:
                raise Exception('Mode must be "LEFT", "CENT", or "RIGH".')
            yield dev.write('TIM:REF {}'.format(side))
        resp = yield dev.query('TIM:REF?')
        returnValue(resp)
```

agilent_infiniium_scope.py
BrooksCampbell: I think I got this change right...

agilent_infiniium_scope.py
BrooksCampbell: done

agilent_infiniium_scope.py
```diff
@@ -409,67 +438,69 @@ def _parsePreamble(preamble):
...
+
+    # Get rid of header
+    # unpack binary data
+    if word_length == 1:
```
maffoo: Here, we should actually read the header to determine if we got enough data. Also, there's no need to have different logic for each word_length. And, I would suggest at the top just doing `import struct` and then using qualified function references here: 
```python
if data[0] != '#':
    raise Exception("Invalid wave data. Expected '#' at start. {}".format(data))
len_header = int(data[1])
wave_offset = 2 + len_header
len_wave = int(data[2:wave_offset])
expected_len = 1 + 1 + len_header + len_wave + 1
if len(data) != expected_len:
    raise Exception("Invalid wave data. Expected {} bytes but got {}. {}".format(expected_len, len(data), data))
wave_data = data[wave_offset:wave_offset+len_wave]
num_words = len_wave // word_length
words = struct.unpack('>' + format_char * num_words, wave_data)
return np.array(words)
```

agilent_infiniium_scope.py
```diff
@@ -409,67 +438,69 @@ def _parsePreamble(preamble):
...
-    return (preambleDict)
+    preample_dict['xUnit'] = unit_type(preample_dict['xUnits'])
+    preample_dict['yUnit'] = unit_type(preample_dict['yUnits'])
+    return (preample_dict)
+
 
-def _parseBinaryData(data, wordLength):
+def _parseBinaryData(data, word_length):
     """Parse binary data packed as string of RIBinary
     """
```
maffoo: Would be good to include a link to the discussion in the manual of the data format: http://www.keysight.com/upload/cmc_upload/All/9000_series_prog_ref.pdf?&cc=US&lc=eng#page=1091

agilent_infiniium_scope.py
BrooksCampbell: fixed

agilent_infiniium_scope.py
BrooksCampbell: I see.   I also caught that I had actually started changing that and didnt finish.  It should be if side is not None.

I refactored horiz_refpoint, trigger_sweep, channelOnOff, trigger_mode, and trigger_edge_slope with this logic change.

agilent_infiniium_scope.py
BrooksCampbell: Fixed.

agilent_infiniium_scope.py
BrooksCampbell: Wahoo!  I finally noticed something!

agilent_infiniium_scope.py
BrooksCampbell: done.

agilent_infiniium_scope.py
BrooksCampbell: done.

agilent_infiniium_scope.py
BrooksCampbell: I might recommend not printing the actual data in the exception.  The expected byte string is many tens of thousands of bytes, so you'll surely miss the important part of the message if you print it all.

Is "//" a real thing or a typo for "/" here:

`num_words = len_wave // word_length`

maffoo: Good point about the data. In the first error, you might want to include just the first byte. In the second, probably just including the length is enough.

`//` _is_ a real thing. It means do integer division, as opposed to floating point division. In python 2 if you divide two integers with `/` you always get an int, but in python 3 dividing `/` gives a float and you have to use `//` to get integer division. You can turn on the python 3 behavior by doing `from __future__ import division`, which we've gotten in the habit of doing for new code. But in any case, `//` does integer division with or without that import, so it's a good habit to get into :-)

agilent_infiniium_scope.py
```diff
@@ -31,446 +31,344 @@
...
+    if data[0] != '#':
+        raise Exception("Invalid wave data. Expected '#' "
+                        "at start. {}".format(data[0]))
+    len_header = int(data[1])
+    wave_offset = 2 + len_header
+    len_wave = int(data[2:wave_offset])
+    expected_len = 1 + 1 + len_header + len_wave + 1
+    if len(data) != expected_len:
+        raise Exception("Invalid wave data. Expected {} bytes "
+                        "but got {}. {}".format(expected_len, len(data)))
```
maffoo: There's one extra `{}` slot in the format string.

agilent_infiniium_scope.py
```diff
@@ -31,446 +31,344 @@
...
-        dat = numpy.array(unpack('>' + formatChar*(len(dat)/wordLength),dat))      
-    return dat
+    format_chars = {1: 'b', 2: 'h', 4: 'f'}
+    format_char = format_chars[word_length]
+
+    # Get rid of header
+    # unpack binary data
+    if data[0] != '#':
+        raise Exception("Invalid wave data. Expected '#' "
+                        "at start. {}".format(data[0]))
```
maffoo: might say `Expected '#' at start but got {}.".format(data[0])`

agilent_infiniium_scope.py
```diff
@@ -31,446 +31,344 @@
...
-        elif factor in probeFactors:
-            yield dev.write(chString+'PRO %d' %factor)
-            resp = yield dev.query(chString+'PRO?')
-        else:
-            raise Exception('Probe attenuation factor not in '+str(probeFactors))
-        returnValue(resp)
-
-    @setting(114, channel = 'i', state = '?', returns = 's')
-    def channelOnOff(self, c, channel, state = None):
+    @setting(114, channel='i', state='?', returns='s')
```
maffoo: `state` must be a string or int, so could say `state=['s', 'w', 'i']`. No reason to give the freedom to pass anything here.

agilent_infiniium_scope.py
```diff
@@ -31,446 +31,344 @@
...
         returnValue(resp)
 
-    @setting(150, side = 's', returns = ['s'])
-    def horiz_refpoint(self, c, side = None):
-        """Get or set the reference point for the horizontal position. Must be 'LEFT', 'CENT'er, or 'RIGH't.
+    @setting(150, side='s', returns=['s'])
+    def horiz_refpoint(self, c, side=None):
+        """Get or set the reference point for the horizontal position.
+
+        Must be 'LEFT', 'CENT'er, or 'RIGH't.
```
maffoo: This might be confusing to people. I would just say `'LEFT', 'CENT', or 'RIGH'`.

agilent_infiniium_scope.py
```diff
@@ -31,446 +31,344 @@
...
-        numPoints = int(preambleDict['numPoints'])
-        time = numpy.linspace(float(preambleDict['xFirst']),
-                              (numPoints-1) * float(preambleDict['xStep'])+float(preambleDict['xFirst']),
-                              numPoints)
-
-        returnValue((time*U.ns*1e9, traceVolts*U.V))
+        # Transfer waveform data
+        p = dev._packet().write('WAV:DATA?').read_raw()
+        result = yield p.send()
+        binary = result['read_raw'][:-1]  # drop termination character
```
maffoo: I just realized that the new implementation of `_parseBinaryData` that I suggested expects this termination character to be there when it looks at the total data length. So no longer need to drop it here.

agilent_infiniium_scope.py
```diff
@@ -31,446 +31,344 @@
...
-        dat = numpy.array(unpack('>' + formatChar*(len(dat)/wordLength),dat))      
-    return dat
+    format_chars = {1: 'b', 2: 'h', 4: 'f'}
+    format_char = format_chars[word_length]
+
+    # Get rid of header
+    # unpack binary data
+    if data[0] != '#':
+        raise Exception("Invalid wave data. Expected '#' "
+                        "at start. Expected '#' but got {}".format(data[0]))
```
maffoo: Just `"at start but got {}".format(data[0]))`. No need to repeat expected value.

agilent_infiniium_scope.py
BrooksCampbell: Yeah, I agree.  I had debated using the more informative string names throughout, but ultimately did not.  i.e. the device supports commands like ":WAVform:PREamble?" or ":WAV:PRE?" but the comment you just brought up is (I believe) the only place in this server that this is mentioned.

er, and t removed.

maffoo:
I just have a couple of small remaining comments, but I think this is coming along nicely. This is a huge improvement over what we had before, so thanks for working on this! With all these changes, especially parsing the binary data, it would be good to test this out. Can you run it on some real hardware with the new gpib server?

agilent_infiniium_scope.py
BrooksCampbell: Got it.  This also menas that I can delete the exception if 'state' isn't an int or string, right?

maffoo: In fact, there's no reason to convert state to a string if it is an int, since it will still get formatted just fine. So can do:
```python
if state is not None:
    if isinstance(state, str):
        state = state.upper()
    if state not in [0, 1, 'ON', 'OFF']:
        raise Exception(...)
    yield dev.write(...)
```

agilent_infiniium_scope.py
BrooksCampbell: fixed

agilent_infiniium_scope.py
BrooksCampbell: Good catch.

agilent_infiniium_scope.py
```diff
@@ -31,446 +31,346 @@
...
-        lenHeader = int(data[1])
-        dat = data[(2+lenHeader):]
-        dat = dat[-calcsize('>' + formatChar*(len(dat)/wordLength)):]
-        dat = numpy.array(unpack('>' + formatChar*(len(dat)/wordLength),dat))      
-    return dat
+    format_chars = {1: 'b', 2: 'h', 4: 'f'}
+    format_char = format_chars[word_length]
+
+    # Get rid of header
+    # unpack binary data
```
maffoo: can probably kill these comments

agilent_infiniium_scope.py
```diff
@@ -17,7 +17,7 @@
 ### BEGIN NODE INFO
 [info]
 name = Agilent Infiniium Oscilloscope
-version = 0.2
+version = 0.2.1
```
maffoo: Given the scope of these changes (no pun intended), I would suggest making this `0.3.0`.

maffoo:
Two small comments, then LGTM from a code perspective. Please test with a scope before merging. And thanks again, this is great!

BrooksCampbell:
I tested through all the functions above get_trace so far and caught some minor things.  I haven't figured out what I broke here though.
```
In [80]: scope.get_trace(1)
---------------------------------------------------------------------------
Error                                     Traceback (most recent call last)
<ipython-input-80-f55e13445fe5> in <module>()
----> 1 scope.get_trace(1)

C:\pylabrad\labrad\client.pyc in __call__(self, *args, **kw)
     65                           "Use setting.future(...) instead.")
     66         f = self.future(*args, **kw)
---> 67         return f.result() if wait else f
     68
     69     def future(self, *args, **kw):

C:\pylabrad\labrad\concurrent.pyc in result(self, timeout)
     98     def result(self, timeout=None):
     99         if not self.__done:
--> 100             self.__result = super(MutableFuture, self).result(timeout)
    101             self.__done = True
    102         if self.__error is not None:

C:\Python27\lib\site-packages\concurrent\futures\_base.pyc in result(self, timeo
ut)
    403                 raise CancelledError()
    404             elif self._state == FINISHED:
--> 405                 return self.__get_result()
    406             else:
    407                 raise TimeoutError()

C:\Python27\lib\site-packages\concurrent\futures\_base.pyc in __get_result(self)

    355     def __get_result(self):
    356         if self._exception:
--> 357             raise type(self._exception), self._exception, self._tracebac
k
    358         else:
    359             return self._result

Error: (0) [Agilent Infiniium Oscilloscope] Remote Traceback (most recent call l
ast):
  File "C:\pylabrad\labrad\server.py", line 232, in request_handler
    result = yield setting.handleRequest(self, c.data, flat_data)
  File "C:\Python27\lib\site-packages\twisted\internet\defer.py", line 1099, in
_inlineCallbacks
    result = g.send(result)
  File "agilent_infiniium_scope.py", line 307, in get_trace
    # Transfer waveform data
  File "agilent_infiniium_scope.py", line 330, in _parsePreamble
    def _parsePreamble(preamble):
exceptions.NameError: global name 'preamble_dict' is not defined
 [payload=None]
```

BrooksCampbell:
Oh, it looks like I just have to initialize preamble_dict before setting the values like this?:

```
    preamble_dict = {}
    for (key, included), val in zip(PREAMBLE_KEYS, preamble_vals):
        if included:
            preamble_dict[key] = val
```

BrooksCampbell:
Alrighty, everything seems to be working well now.  I added some exceptions to make sure channel was in [1,2,3,4] in a few places, and I also added a check in get_trace to make sure that the requested channel was enabled.

I pulled data from the device and plotted it and it seems to be scaled correctly and everything.

All good to merge?

BrooksCampbell:
ugh, it seems I accidentally pulled the gpib_server changes from master into this branch as well... do I have to figure out how to pull that back out, or can it just be re-merged with master?

agilent_infiniium_scope.py
```diff
@@ -31,446 +31,354 @@
...
-            else:
-                raise Exception('state must be int or string')
-            if state not in ['0','1','ON','OFF']:
-                raise Exception('state must be 0, 1, "ON", or "OFF"')
-            yield dev.write(('CHAN%d:DISP '+state) %channel)
-            resp = yield dev.query('CHAN%d:DISP?' %channel)
+            if state not in [0, 1, 'ON', 'OFF']:
+                raise Exception('state must be 0, 1, "ON", or "OFF". '
+                                'Got {}'.format(state))
+            if channel not in [1, 2, 3, 4]:
```
maffoo: This check should go above, outside the `if` statement, since you use the channel in the query later, even if `state` is `None`.

maffoo: There are a number of other settings that take `channel` as an argument that would benefit from the same check. It would be good to factor the check out into a separate method that can be reused. However, I'd suggest doing that in a separate commit since this one is already pretty big :-)

maffoo:
I would suggest pulling out the gpib changes from the branch; I'm not sure how github will merge those. It is reporting that there are no conflicts with master, but still showing the files as changed, which is odd and makes me worried. `git rebase -i` should be able to fix this up easily enough; the pyle contributing file describes how to do this if you're unfamiliar.

BrooksCampbell:
Jimmy just helped me rebase.  Did you want a separate pull request for the channel string check, or just a separate commit in this pull request?

maffoo:
By "channel string check" do you mean the `_check_channel` function, or something else? It's find to include that function in this PR, so if that's all you meant them LGTM.

PR #371, opened 2016-08-31
Added write_raw to gpib_server.py

dahlend:
I had a typo in my original pull request.


gpib_server.py
```diff
@@ -165,6 +165,11 @@ def write(self, c, data):
         """Write a string to the GPIB bus."""
         self.getDevice(c).write(data)
 
+    @setting(8, data='s', returns='')
```
maffoo: In the recent (scala) versions of the manager there is a new type `'y'` for binary data. The intent is that `'s'` is for text strings in utf8 encoding, while `'y'` can contain arbitrary bytes. Note that the `read_raw` setting below returns `'y'`; we should use that type here as well.

maffoo:
Since this is adding a setting, please bump the server version in the node config block at the top to `1.5.0-no-refresh`.

maffoo:
Please see comments, then I'm happy to merge. Note that you can just push additional commits to your branch and this PR will get updated; no need to open a new PR.

dahlend:
OK, changes made.

maffoo:
Thanks, @dahlend!

PR #373, opened 2016-10-12
Remove file logging in ADR server.

maffoo:
Log messages will still get printed to standard output, and these are in turn logged by the node when the server is run from there. This fixes an issue where the server failed to start because the working directory was not writeable.


BrooksFoxen:
LGTM.

PR #375, opened 2016-10-12
Improve PyQt installation instructions

DanielSank:
@BrooksFoxen @maffoo @ejeffrey


DanielSank:
Actually, I'm wondering if this should be posted on Stack Overflow instead. Thoughts?

DanielSank:
Could someone please verify that this works in a virtualenv? @ejeffrey @zchen088 @BrooksFoxen @maffoo @kunalq?

brooksfoxen:
I just tried this installation procedure on a different virtual environment on the same computer where we developed the procedure.  I started after installing the dev tools since I already had those and it seems to have worked.

```
(pyqt_test)brookscampbell@miramar:~/src/servers$ python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import PyQt4
>>> import sip
>>> quit()
```
Unfortunately I forgot to check this in the same virtualenv before doing the install, but I had just copied pyqt_test from 2.7 as shown below.
```
brookscampbell@miramar:~/src/servers$ cpvirtualenv 2.7 pyqt_test2
Copying 2.7 as pyqt_test2...
(pyqt_test2)brookscampbell@miramar:~/src/servers$ python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import sip
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named sip
>>> import PyQt4
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named PyQt4
```

maffoo:
@DanielSank, another option is to just use the os-provided versions of pyqt:
```
$ sudo apt-get install python-qt4 python-qwt5-qt4
$ mkvirtualenv test --system-site-packages
$ workon test
```
One can still install stuff in this virtualenv that is newer than the system versions, but will fall back to the system versions otherwise. If the versions of pyqt provided by the system are new enough, then this seems like a simple option. Here are the versions I get, following instructions [here](https://wiki.python.org/moin/PyQt/Getting%20the%20version%20numbers%20of%20Qt,%20SIP%20and%20PyQt):
```python
In [1]: from PyQt4.QtCore import QT_VERSION_STR

In [2]: from PyQt4.Qt import PYQT_VERSION_STR

In [3]: from sip import SIP_VERSION_STR

In [4]: QT_VERSION_STR, PYQT_VERSION_STR, SIP_VERSION_STR
Out[4]: ('4.8.6', '4.10.4', '4.15.5')
```

DanielSank:
@maffoo Yeah, that works, but since it's relatively easy to get a virtualenv-specific version, why not use it?

maffoo:
@DanielSank, I think at the very least the docs should include instructions for how to use system versions of pyqt, since that is simple and will probably suffice for most usage (also, this requires O(1) work even for N different virtualenvs). Can include more detailed directions about how to compile and install directly into a virtualenv if the user needs that.

A couple of questions about the install directions:
- why move this into a separate google doc?
- is it important that the sip include directory be somewhere in the virtualenv, or does one just need to use the same directory when compiling sip and pyqt?

DanielSank:
> @DanielSank, I think at the very least the docs should include instructions for how to use system versions of pyqt, since that is simple and will probably suffice for most usage (also, this requires O(1) work even for N different virtualenvs).

One of the problems with using the system-wide solution is that it becomes unclear as to where virtualenv packages are coming from. As you say, versions can be checked, but at least for my part I'd rather not have to think about it at all. Since there's no simple way to tell a virtualenv to take *just* the PyQt system lib, I'd rather just not use the system packages approach at all.

That said, I've copied your instructions to the Google Doc.

> why move this into a separate google doc?

Because no particular git repo seems like the right place and I'd rather just be able to link to the instructions from wherever needed.

> is it important that the sip include directory be somewhere in the virtualenv, or does one just need to use the same directory when compiling sip and pyqt?

It needs to be written to a directory to which we can write without sudo.

maffoo:
> It needs to be written to a directory to which we can write without sudo.

Maybe mention that this directory is arbitrary and does not need to be in the virtualenv; this is just a place that is guaranteed to be user-writeable and won't interfere with installing other versions of pyqt in other virtualenvs.

Also, the `.virtualenvs` directory only exists and the `mkvirtualenv` command is only available if using virtualenvwrapper with the default configuration. Should mention this so users can modify the instructions accordingly if they're just doing plain old virtualenvs.

maffoo:
After that, LGTM.

PR #376, opened 2016-10-13
ADR: clean up handling of peripheral ids and server names.

maffoo:
This fixes a problem with node names that contain characters such as hyphens that get modified by python name mangling. Also cleans up code style and docstrings in modified functions.


ADR/ADR.py
```diff
@@ -1006,34 +1030,41 @@ class ADRServer(DeviceServer):
...
-                    if "node_" + node.lower() not in self.client.servers:
-                        missing_nodes.append(node)
-                if not missing_nodes:
-                    device_list.append((name, (self.client,)))
-                else:
-                    print "device %s missing nodes: %s" % (name, str(list(set(missing_nodes))))
-                yield reg.cd(1)
+        dirs, keys = yield reg.dir()
+        for name in dirs:  # dirs are adr names.
+            if 'defaults' in name:
```
BrooksFoxen: do we want to exclude the 'backups' folder too? i.e. `if 'defaults in name or 'backups' in name:`

Perhaps it doesn't matter because there are no keys in the backups folder, but maybe there would be in the future.

maffoo: Changed to skip 'backups' as well.

PR #377, opened 2016-10-14
Introduce master oscilloscope server.

DanielSank:
Heck yes!

Suggested review strategy:

1. Review `oscilloscope_server.py` as this defines the user interface.

1. Review `wrappers.py`, the base class for oscilloscope wrappers.

1. Review `agilent.py`. There is only one scope implemented so far. All settings have been tested on a real scope.

The interfaces here are not guaranteed to be backward compatible with any of the other scope servers.


oscilloscope/wrappers.py
```diff
@@ -0,0 +1,318 @@
...
+
+    TRIGGER_LEVEL_UNIT = NotImplementedAttribute('TRIGGER_LEVEL_UNIT')
+    trigger_level_write = NotImplementedAttribute('trigger_level_write')
+    trigger_level_query = NotImplementedAttribute('trigger_level_query')
+    trigger_level_parse = NotImplementedAttribute('trigger_level_parse')
+
+    TRIGGER_MODES = ["AUTO", "NORM"]
+    trigger_mode_write = 'TRIG:MOD {0}'.format
+    trigger_mode_query = 'TIRG:MOD?'.format
+    trigger_mode_parse = str  # XXX is this correct?
```
DanielSank: We have not yet implemented trigger modes on any of the scope wrappers.

DanielSank:
@ejeffrey @maffoo @kunalq could one of you guys pick this one up?

oscilloscope/agilent.py
```diff
@@ -0,0 +1,195 @@
...
+    We mostly override VISA string functions from the parent class, but in some
+    cases we have to write complete self-contained methods. This file should be
+    viewed along side the parent class in order to understand what's going on.
+
+    See programmer manual at:
+        http://www.keysight.com/upload/cmc_upload/All/4000_series_prog_guide.pdf?&cc=US&lc=eng
+    """
+
+    # CHANNEL
+
```
BrooksFoxen: I've been sloppy with leading colons (:), but they do have an affect.  I don't exactly understand why you would ever *not* want to send a leading colon (which places you at the [root of the command tree structure](http://www.keysight.com/upload/cmc_upload/All/Infiniium_prog_guide.pdf?&cc=US&lc=eng#page=178).

It seems like you're trading sending fewer characters in your GPIB commands  (if you attempt to not always use a leading colorn) for the potential risk of sending a command while in the wrong tree and having it ignored?

oscilloscope/agilent.py
```diff
@@ -0,0 +1,195 @@
...
+            'x_incriment': x_incriment,
+            'x_origin': x_origin,
+            'x_reference': x_reference,
+            'y_incriment': y_incriment,
+            'y_origin': y_origin,
+            'y_reference': y_reference}
+
+
+def parse_binary_waveform(data, word_length):
+    """Convert binary waveform dump to numpy array.
```
BrooksFoxen: As written, this parse binary waveform function only works when :WAVform:STReaming is off.  We should probably support both.  [Pages 1091-1092 in the Infiniium manual describe both binary block formats](http://www.keysight.com/upload/cmc_upload/All/9000_series_prog_ref.pdf?&cc=US&lc=eng#page=1091).

oscilloscope/agilent.py
```diff
@@ -0,0 +1,195 @@
...
+
+from __future__ import absolute_import
+import oscilloscope.wrappers as wrappers
+import labrad.units as U
+import struct
+import numpy as np
+from twisted.internet.defer import inlineCallbacks, returnValue
+
+
+class AgilentDSOX4104AWrapper(wrappers.OscilloscopeWrapper):
```
BrooksFoxen: When working on the Infiniium Scope Server, I noticed that the scope supports more descriptive GPIB commands.

i.e. sending the scope ":WAV:STR ON" is equivalent to "WAVeform:STReaming ON" or ":TIMebase:REFerence CENTer" etc.  This is a nice form of self-documentation that we might want to start using if other scope's support it.  As another note, writing commands in this form also makes finding them in the scope's documentation easier as the programming guide always writes the commands in this longer format.

oscilloscope/oscilloscope_server.py
```diff
@@ -0,0 +1,227 @@
+"""
```
maffoo: Consider calling this file just `server.py`. The module name `oscilloscope.server` is unambiguous and less repetitive.

oscilloscope/agilent.py
DanielSank: Could we make this an issue to fix after this PR is landed?

oscilloscope/agilent.py
DanielSank: > It seems like you're trading sending fewer characters in your GPIB commands (if you attempt to not always use a leading colorn) for the potential risk of sending a command while in the wrong tree and having it ignored?

It wasn't intentional. Will fix.

oscilloscope/oscilloscope_server.py
```diff
@@ -0,0 +1,227 @@
...
+    @setting(12, returns='')
+    def clear_buffers(self, c):
+        """Clear device status buffers."""
+        dev = self.selectedDevice(c)
+        yield dev.clear_buffers()
+
+    # CHANNEL
+
+    @setting(50, channel='i', state='s', returns='s')
+    def channel_on_off(self, c, channel, state=None):
```
DanielSank: I'm trying to maintain some level of backward compatibility with existing code. This method is unfortunately called `channel_on_off`, which to me does not mix well with boolean input. I can change it, of course, but it seems funky. On the other hand, I could say that this is a new server with a new interface and change the name to `channel_on` with boolean input. Thoughts?

oscilloscope/oscilloscope_server.py
```diff
@@ -0,0 +1,227 @@
...
+        dev = self.selectedDevice(c)
+        result = yield dev.channel_on_off(channel, state)
+        returnValue(result)
```
DanielSank: I forgot you can do that. Thanks.

oscilloscope/oscilloscope_server.py
```diff
@@ -0,0 +1,227 @@
...
+        result = yield dev.invert(channel, invert)
+        returnValue(result)
+
+    @setting(53, channel='i', term='i', returns='i')
+    def termination(self, c, channel, term=None):
+        """Set channel termination
+
+        Args:
+            channel (int): Which channel to set termination.
+            term (int): Termination in Ohms. Either 50 or 1,000,000 on every
```
DanielSank: Definitely not using `v[Ohm]` as floating points don't make sense here. The string version could work, and perhaps makes more sense.

oscilloscope/oscilloscope_server.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+        dev = self.selectedDevice(c)
+        resp = yield dev.trigger_level(level)
+        returnValue(resp)
+
+    # ACQUISITION
+
+    @setting(60, channel='i', returns='*v[s]*v[V]')
+    def get_trace(self, c, channel):
+        """Get a trace for a single channel.
```
DanielSank: > Is the idea that this downloads a previously-captured trace, or does it capture and download?

The "idea" here (if you can call it that) was to replicate how the previous scope servers work so I can do work. We should set up nice interfaces in the future, and maybe we should rename this one to make room for future improvements, but I'm not at all interested in adding new functionality now.

> I don't really see any settings here for doing things like triggering a single-shot capture; do you plan to support that sort of thing in a generic way?

oscilloscope/wrappers.py
```diff
@@ -0,0 +1,318 @@
...
+    # input
+
+    CHANNEL_STATES = ['ON', 'OFF']
+    channel_on_off_write = NotImplementedAttribute('channel_on_off_write')
+    channel_on_off_query = NotImplementedAttribute('channel_on_off_query')
+    channel_on_off_parse = NotImplementedAttribute('channel_on_off_parse')
+
+    COUPLINGS = ['AC', 'DC']
+    coupling_write = 'CH{0}:COUP {1}'.format
+    coupling_query = 'CH{0}:COUP?'.format
```
DanielSank: The reason was a failed attempt to make some of the strings work on both Tek and Agilent devices.

oscilloscope/oscilloscope_server.py
DanielSank: done

oscilloscope/agilent.py
DanielSank: I think I got each case.

oscilloscope/agilent.py
DanielSank: I agree but let's do this as a separate step.

oscilloscope/agilent.py
```diff
@@ -0,0 +1,195 @@
...
+
+"""
+
+
+from __future__ import absolute_import
+import oscilloscope.wrappers as wrappers
+import labrad.units as U
+import struct
+import numpy as np
+from twisted.internet.defer import inlineCallbacks, returnValue
```
DanielSank: I'm really glad to know I'm not the only person who twitches when imports aren't organized.

Fixed.

oscilloscope/oscilloscope_server.py
```diff
@@ -0,0 +1,227 @@
...
+### END NODE INFO
+"""
+
+from __future__ import absolute_import
+from labrad.server import setting
+from labrad.gpib import GPIBManagedServer
+from twisted.internet.defer import inlineCallbacks, returnValue
+
+
+import oscilloscope.tektronix as tektronix
```
DanielSank: fixed

oscilloscope/oscilloscope_server.py
DanielSank: Ok, switched to boolean.

oscilloscope/oscilloscope_server.py
DanielSank: Done.

oscilloscope/oscilloscope_server.py
```diff
@@ -0,0 +1,227 @@
...
+        returnValue(result)
+
+    @setting(52, channel='i', invert='i', returns='i')
+    def invert(self, c, channel, invert=None):
+        """Get or set channel inversion.
+
+        Args:
+            channel (int):
+            invert (int): 1 -> invert channel, 0 -> do not invert channel. If
+                None (the default), then we only query the inversion.
```
DanielSank: Fixed.

oscilloscope/oscilloscope_server.py
```diff
@@ -0,0 +1,227 @@
...
+        Returns:
+            (float): The horizontal position in units of divisions.
+        """
+        dev = self.selectedDevice(c)
+        result = yield dev.horiz_position(position)
+        returnValue(result)
+
+    # TRIGGER
+
+    @setting(71, source=['s'], returns='s')
```
DanielSank: fixed

oscilloscope/oscilloscope_server.py
```diff
@@ -0,0 +1,227 @@
...
+        Returns:
+            (ValueArray[s]): Time axis.
+            (ValueArray[V]): Voltages.
+        """
+        dev = self.selectedDevice(c)
+        result = yield dev.get_trace(channel)
+        returnValue(result)
+
+
+__server__ = OscilloscopeServer()
```
DanielSank: fixed

oscilloscope/wrappers.py
```diff
@@ -0,0 +1,318 @@
+from __future__ import absolute_import
+
+from labrad import types as T
+from labrad.gpib import GPIBDeviceWrapper
+from twisted.internet.defer import inlineCallbacks, returnValue
+
+import numpy
```
DanielSank: fixed

oscilloscope/wrappers.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+def require_is_in(obj, collection):
+    if obj not in collection:
+        raise ValueError("%s not in %s"%(obj, collection))
```
DanielSank: fixed

oscilloscope/wrappers.py
```diff
@@ -0,0 +1,318 @@
...
+        'CH0:COUP ON'
+    which is the VISA command that turns on channel 0.
+    """
+
+    # CHANNEL
+
+    # input
+
+    CHANNEL_STATES = ['ON', 'OFF']
+    channel_on_off_write = NotImplementedAttribute('channel_on_off_write')
```
DanielSank: done

oscilloscope/wrappers.py
```diff
@@ -0,0 +1,318 @@
...
+    channel_on_off_query = NotImplementedAttribute('channel_on_off_query')
+    channel_on_off_parse = NotImplementedAttribute('channel_on_off_parse')
+
+    COUPLINGS = ['AC', 'DC']
+    coupling_write = 'CH{0}:COUP {1}'.format
```
DanielSank: done

oscilloscope/wrappers.py
```diff
@@ -0,0 +1,318 @@
...
+
+    # CHANNEL
+
+    @inlineCallbacks
+    def channel_on_off(self, channel, state=None):
+        """Get or set channel on/off.
+
+        Args:
+            channel - int: The channel to get or set.
+            state - str: 'ON' or 'OFF'.
```
DanielSank: I think in converting some stuff to boolean, we're covered everywhere now.

DanielSank:
@maffoo  please see 055b7727c38a69df1371c527c9d6c28032fd06ba in which we use method factories to reduce granularity of the VISA command parametrizations and to reduce code replication.

DanielSank:
@maffoo bump

DanielSank:
@maffoo bump

oscilloscope/agilent.py
brooksfoxen: LGTM.  

Note, the "...you're trading..." was a general 'you' and not directed at you personally.  =D

oscilloscope/wrappers.py
```diff
@@ -0,0 +1,293 @@
...
+    trigger_slope_parse = nia('trigger_slope_parse')
+
+    TRIGGER_LEVEL_UNIT = nia('TRIGGER_LEVEL_UNIT')
+    trigger_level_write = nia('trigger_level_write')
+    trigger_level_query = nia('trigger_level_query')
+    trigger_level_parse = nia('trigger_level_parse')
+
+    TRIGGER_MODES = ["AUTO", "NORM"]
+    trigger_mode_write = ':TRIG:MOD {}'.format
+    trigger_mode_query = ':TIRG:MOD?'.format
```
brooksfoxen: Typo ":TIRG:MOD?" should be ":TRIG:MOD?"

maffoo:
@DanielSank, yes, the `channel_method` and `global_method` functions look great! I would suggest calling the first and last params `parse_input` and `parse_output` instead, but up to you. Also, note that `NotImplementedError` is an exception _class_, so you should be doing `raise NotImplementedError()`.

ejeffrey:
@maffoo: it is actually OK to raise an exception class rather than an instance.  The exception machinery will create the instance for you.  Yay python and lazy typing!  But I agree that we should use the instance form since you need to do that if you want to pass an error message.

DanielSank:
>  But I agree that we should use the instance form since you need to do that if you want to pass an error message.

@ejeffrey Huh! I did not know that.

DanielSank:
@maffoo I think we're all done now. Just waiting for your LGTM.

maffoo:
LGTM

PR #388, opened 2016-11-21
Added functions for the complete set of configuration parameters.

brooksfoxen:
Some functions were missing from the SR830 server.  I added functions for all the configuration parameters so that the instrument can be fully configured into a known state.


SR830/SR830.py
```diff
@@ -401,8 +426,124 @@ def output_overload(self, c):
...
+        """
+        coupling_dict = {
+            'AC':0,
+            'DC':1
+        }
+        dev = self.selectedDevice(c)
+        if coupling is not None:
+            if isinstance(coupling, str):
+                coupling = coupling.upper()
+                if (coupling not in coupling_dict.keys()) and (coupling not in [0,1]):
```
thchwhite: can replace this line with:

If coupling not in (coupling_dict.keys() + [0,1]) to make it shorter

SR830/SR830.py
```diff
@@ -401,8 +426,124 @@ def output_overload(self, c):
...
+
+        Args:
+            mode:  0: High Reserve, 1: Normal, 2: Low Noise
+        Returns:
+            (int):  The reserve mode after setting (if requested).
+        """
+        dev = self.selectedDevice(c)
+        if mode is not None:
+            mode = int(mode)
+            if mode not in [0,1,2]:
```
thchwhite: spaces

SR830/SR830.py
```diff
@@ -401,8 +426,124 @@ def output_overload(self, c):
...
+        """Get or sets the input coupling.
+
+        Args:
+            coupling:  0: AC or 1: DC
+        Returns:
+            (int):  The input coupling after setting (if requested).
+        """
+        coupling_dict = {
+            'AC':0,
+            'DC':1
```
thchwhite: nit: spaces between key and value

SR830/SR830.py
thchwhite: also spaces between 0 and 1

SR830/SR830.py
```diff
@@ -61,6 +61,14 @@ def getSensitivity(i):
     else:
         return 10 * 10**(-9 + i/3)
 
+MODE_DICT = {
```
thchwhite: nit: spaces between key and value

PR #394, opened 2016-12-27
telecommServer improvements

amopremcak:
Added some instructions for setting up the registry information required to use this server. Added an optional password argument to the send_sms setting for cases in which authentication is required. Changes should be backward compatible.


Telecomm/telecommServer.py
```diff
@@ -32,6 +32,25 @@
...
+#       ,("USERNAMEM","TENDIGITPHONENUMBER@vtext.com")]
+#      Note that the domain name depends on the user's
+#      mobile phone provider. "USERNAME" must be ALL caps.
+#   3) Create a key named 'domain' with a value that
+#      matches the domain name of the email address
+#      from which you wish to send messages.
+#      e.g. domain = "physics.someschool.edu"
+#   4) Create a key named 'smtpServer' whose value is
+#      the smtp server you wish to use.
+#   5) (Optional) Create a key named 'password'
```
DanielSank: I wonder if there's a better place to store the password.

Telecomm/telecommServer.py
```diff
@@ -127,27 +150,35 @@ def _refreshConnectionData(self):
         reg = cxn.registry
         p = reg.packet()
         p.cd(REGISTRY_PATH)
+        p.dir(key = 'regcontent')
```
DanielSank: No space around `=` on keyword args.

Telecomm/telecommServer.py
```diff
@@ -127,27 +150,35 @@ def _refreshConnectionData(self):
...
         p.get(SMS_KEY, key='userlist')
         p.get(DOMAIN_KEY, key='domain')
         p.get(SERVER_KEY, key='server')
         resp = yield p.send()
         self.smsUsers=dict(resp['userlist'])
         self.domain = resp['domain']
         self.smtpServer = resp['server']
+        regContent = resp['regcontent'][1]
```
DanielSank: `keys` might be a better variable name.

Telecomm/telecommServer.py
```diff
@@ -127,27 +150,35 @@ def _refreshConnectionData(self):
...
             
-    @setting(11, subject='s', msg='s', recipients=['*s','s'], returns='b{success}*s{failures}')
-    def send_sms(self, c, subject, msg, recipients):
+    @setting(11, subject='s', msg='s', recipients=['*s','s'], username='s', returns='b{success}*s{failures}')
+    def send_sms(self, c, subject, msg, recipients, username='LabRAD'):
         if not isinstance(recipients,list):
             recipients = [recipients]
         recipients = [self.smsUsers[name.upper()] for name in recipients if name.upper() in self.smsUsers]
-        success, failures = textMessage(recipients, subject, msg, self.domain, self.smtpServer)
+        success, failures = textMessage(recipients, subject, msg, self.domain, self.smtpServer, username, self.password)
```
DanielSank: Long line?

amopremcak:
All of the requested changes are complete @DanielSank

Telecomm/telecommServer.py
amopremcak: I like the registry more than a plain text file, because one has to jump through some additional loops in order to find it.

DanielSank: Well, the difference is that the registry is accessible over LabRAD, so putting a password in the registry means that everyone allowed on LabRAD is allowed to see that password. LabRAD isn't really set up to manage permissions like that, so I think it unadvised to store passwords in the registry.

Can you just put in in a yaml or text file and have the server load it?

@maffoo @kunalq thoughts?

kunalq: I'm inclined to agree with @DanielSank -- storing a password in the registry isn't ideal.

amopremcak: If the concern is managing permissions, then would storing the password in a text or yaml file really improve things? Do you guys have a single machine dedicated to using this server? In that case you could create some password file, store it in Telecomm folder and just make sure it is in the git ignore file.

DanielSank: > If the concern is managing permissions, then would storing the password in a text or yaml file really improve things?

Yes. Presumably your computers have access control.

> Do you guys have a single machine dedicated to using this server?

No, but that's not really the point. The point is to not use LabRAD as an access control system, because it's not (yet) designed for that (although @maffoo has been making great strides in that direction).

It's better to distribute the passwords some other way; whether by talking to each other or using some other software tool is irrelevant as long as it's as secure. This leaves the question of how to get the telecomm server to find out about this password.

We recently had to deal with a very similar situation regarding access to an SQL database through some scripts in our own code repo. My first solution was exactly what you're suggesting: put a password file in the same repo but add it to `.gitignore` so it doesn't wind up in version control. Again, how you tell other what password to put in that file is entirely up to you. However, that was a bit limiting and kind of dangerous, so now we have users put the passwords in a file in their home directory, and our scripts simply look there. You can use `os.path.expanduser` to get the home directory on Mac/Linux/Windows.

amopremcak: When I was working on database stuff where I used to work, we transitioned from storing username/password in plain text to windows authentication which is pretty nice, but everyone was running windows 7. Also, we were using MS SQL Server so I don't know if windows authentication was even an option for you guys.

Let me know what you guys want to do with the password stuff.

DanielSank: We certainly need a cross-platform solution. I suppose one option would be to put a key in the registry indicating *how* to get the password, i.e. via a file at a particular location, or fro Windows authentication, etc.

PR #402, opened 2017-01-17
U/brooks/add 34461 a dmm

brooksfoxen:
Add support for the new DMM model, units, and current measurements.

This adds support for the new Agilent 34461A DMM, adds labrad units to returned values, and adds the configure and read current functions to add functionality (the existing "current" function doesnt allow you to set the range or coupling--every time you call "current" those settings are overwritten with defaults.)


agilent_34401A_dmm.py
```diff
@@ -35,18 +35,32 @@
 from labrad.server import setting
 from labrad.gpib import GPIBManagedServer, GPIBDeviceWrapper
 from twisted.internet.defer import inlineCallbacks, returnValue
+from labrad.units import V, mV, Ohm, A, mA, Hz
+import time
+
+AC_COUPLING = {
+    True:'AC',
+    False:'DC'
+}
```
thchwhite: spaces in between key and value

agilent_34401A_dmm.py
```diff
@@ -76,7 +90,51 @@ def read_voltage(self, c):
...
         ans = yield dev.query("READ?")
-        returnValue(float(ans))
+        returnValue(float(ans) * V)
+
+
+    @setting(15, AC = 'b', ac_highpass='v[Hz]', cRange='v[A]',
+             resolution ='v[V]')
+    def configure_current(self, c, AC=False, ac_highpass=3*Hz,
+                          cRange='AUTO', resolution=0.0001):
+        """ Sets the DMM to current mode, with given range and resolution. """
```
thchwhite: add documentation for function.

thchwhite: Also resolution default value is not v[V], and it doesn't appear to be used in the function.

thchwhite:
I left a few comments.  Most are nits, but one is important in configure_current.  Also you should fix the branch file name conflict.  Is it 34461 or 34401?

brooksfoxen:
I made changes--as for the 34401 and 34461 issue.  Keysight has released a new version of the 34401A, the 34461A.  They're very similar DMMs, and potentially identical as far as the very limited set of functions implemented in this server.

It seems like probably we should change the name of the whole file/server to just "DMM" or something like that, but I didn't want to break other people's code (not that it seems like this server gets much use).

PR #405, opened 2017-01-20
Turns tls_mode off in ADR gui to prevent labrad timeouts.

thchwhite:
This seems to be the only way to currently resolve the problem.  I'm open to a better solution if it exists.


thchwhite:
This was actually just an issue with my choice of domain name

PR #406, opened 2017-02-01
uses read_raw function to fix intermittent white space GPIB error.

thchwhite:
Finally tracked down that intermittent error on the spectrum analyzer where it would fail one time out of 100 because of a lack on newline characters.


DanielSank:
@maffoo oh man, good catch.

PR #418, opened 2017-08-07
Move code into labrad.servers package

maffoo:
This fixes up the package structure of this repo so that it can be put on the python path without polluting the global package namespace. Mostly file renames, but also had to update some imports for modules that refer to each other. For these imports, I used absolute imports starting with `labrad.servers`


maffoo:
Gonna close this for now; we can bring in other projects without changing the directory structure in this way.

maffoo:
I added `labrad/servers/__init__.py` with the magic snippet to make this a namespace package, in case the import order is different. From my local testing, it seems this gets picked up by the import machinery and it is not necessary to also add `labrad/__init__.py` in this repo.

ejeffrey:
I believe the "real" answer is that whichever package is first in the users PYTHONPATH gets imported first, and needs to have the magic pkgutil snippet to make the later entries get searched.

